PR #7, opened 2021-07-23
(0.5.6 Device Drivers) unix -> Unix & by a comma? -> by a comma.

alan23273850:


Makefile
```diff
@@ -1,5 +1,5 @@
 all: lkmpg.tex
-	pdflatex -shell-escap lkmpg.tex
+	pdflatex -shell-escape lkmpg.tex
```
jserv, line 2: Don't touch `Makefile`. Minimize the functional changes.

jserv (changes requested):
The subject was misleading. In fact, the change was not meant to be eliminate the typo.

Makefile
alan23273850: I've already discarded this change!

alan23273850:
> The subject was misleading. In fact, the change was not meant to be eliminate the typo.

OK, I'll be aware of it in the future.

jserv:
Thank @alan23273850 for contributing!

PR #9, opened 2021-07-25
Add header and footer

chses9440611:


jserv:
Thank @chses9440611 for contributing!

jserv:
This commit was dropped because it breaks the table of content. Please think of appropriate ways to enable both TOC and header/footer.

PR #11, opened 2021-07-25
Fix command style

2011eric:
Little changes regarding the use of `sudo`.

### DIFF
--- a/lkmpg.tex
+++ b/lkmpg.tex
@@ -400,7 +400,7 @@ It takes two parameters: a variable name and a free form string describing that 
 
 I would recommend playing around with this code:
 \begin{code}
-# sudo insmod hello-5.ko mystring="bebop" myintArray=-1
+# insmod hello-5.ko mystring="bebop" myintArray=-1
 myshort is a short integer: 1
 myint is an integer: 20
 mylong is a long integer: 9999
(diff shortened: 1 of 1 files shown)


jserv:
I would like to enforce the assumption of using as a non-root user. That is, it is necessary to specify `sudo` for the essential commands such as `insmod` and `rmmod`. In addition, most command prompts should start with `$` (dollar sign).

2011eric:
I've specified `sudo` command and replaced `#` with `$`.

jserv (changes requested):
Rebase the latest `master` branch.

jserv:
You should manipulate with not only this block but also other starting with `\begin{verbatim}`. e.g.,
```
\begin{verbatim}
# cat /proc/helloworld
HelloWorld!
\end{verbatim}
```

Change the leading `#` into `$`.

jserv:
After rebasing, you should force push.

2011eric:
I rebased, added a commit , and force pushed. Will that be a problem?

jserv:
> I rebased, added a commit , and force pushed. Will that be a problem?

You only force push your own branch rather than master branch, which does not hurt anything.

jserv:
Use `git rebase -i` to drop commit "Add header and footer"

jserv:
Thank @2011eric for contributing. I amended the commit message for precise wording.

Issue #6, opened 2021-07-23
The `proc_ops` structure should be `file_operations`.


asas1asas200:
Chapter 0.6.1 introduce the `proc_ops` structure and it says that it is defined in `/urs/include/linux/fs.h`, but the `fs.h` never defines this structure since kernel 2.6.12rc2 according to this:
https://github.com/torvalds/linux/blame/master/include/linux/fs.h#L2022
According to the above, I think this structure should be `file_operations`.

I also find the `proc_ops` structure in this:
https://github.com/torvalds/linux/commit/d56c0d45f0e27f814e87a1676b6bdccccbc252e9
It applied since kernel 5.6-rc1 to improve VFS performance.

jserv:
Good. Please send pull request(s) for API transition.

PR #13, opened 2021-07-25
fix(0.6-proc_ops): changed `proc_ops` to `file_operations`

asas1asas200:
Changed `proc_ops` to `file_operations` and also updated its definitions to kernel 5.4 because chap.0.2 install that version.

But I have a question about later chapter.
In chapter 0.7 introduces the /proc file system, use `proc_ops` structure in /proc is fine but that is defined in `linux/proc_fs.h` and you haven't explain it so I may confuse about what is that and the differences between `proc_ops` and `file_operations`.

Related issue: #6

### DIFF
--- a/lkmpg.tex
+++ b/lkmpg.tex
@@ -813,7 +813,7 @@ The meaning is clear, and you should be aware that any member of the structure w
 
 An instance of struct \verb|file_operations| containing pointers to functions that are used to implement read, write, open, \ldots{} syscalls is commonly named fops.
 
-In Linux v5.6+, this structure has been instead by \ref{sec:proc_ops} in the \textbf{/proc} file system.
+Sin Linux v5.6, the \verb|proc_ops| structure was introduced to replace the use of the \verb|file_operations| structure when registering proc handlers.
 
 \subsection{The file structure}
 \label{sec:file_struct}
@@ -954,7 +954,7 @@ HelloWorld!
 \subsection{The proc\_ops Structure}
 \label{sec:proc_ops}
 The \verb|proc_ops| structure is defined in \textbf{/usr/include/linux/proc\_fs.h} in Linux v5.6+.
-In old kernel, it used \verb|file_operations| for custom hooks in \textbf{/proc} file system, but it contains some members that are unnecessary in VFS, and every time VFS expands \verb|file_operations| set, \textbf{/proc} code comes bloated.
+In older kernels, it used \verb|file_operations| for custom hooks in \textbf{/proc} file system, but it contains some members that are unnecessary in VFS, and every time VFS expands \verb|file_operations| set, \textbf{/proc} code comes bloated.
 On the other hand, not only the space, but also some operations were saved by this structure to improve its performance.
 For example, the file which never disappears in \textbf{/proc} can set the \textbf{proc\_flag} as \textbf{PROC\_ENTRY\_PERMANENT} to save 2 atomic ops, 1 allocation, 1 free in per open/read/close sequence.
 
--- a/lkmpg.tex
+++ b/lkmpg.tex
@@ -770,12 +770,12 @@ struct file_operations {
     long (*fallocate)(struct file *file, int mode, loff_t offset,
         loff_t len);
     void (*show_fdinfo)(struct seq_file *m, struct file *f);
-  ssize_t (*copy_file_range)(struct file *, loff_t, struct file *,
-      loff_t, size_t, unsigned int);
-  loff_t (*remap_file_range)(struct file *file_in, loff_t pos_in,
-           struct file *file_out, loff_t pos_out,
-           loff_t len, unsigned int remap_flags);
-  int (*fadvise)(struct file *, loff_t, loff_t, int);
+    ssize_t (*copy_file_range)(struct file *, loff_t, struct file *,
+        loff_t, size_t, unsigned int);
+    loff_t (*remap_file_range)(struct file *file_in, loff_t pos_in,
+             struct file *file_out, loff_t pos_out,
+             loff_t len, unsigned int remap_flags);
+    int (*fadvise)(struct file *, loff_t, loff_t, int);
 } __randomize_layout;
 \end{code}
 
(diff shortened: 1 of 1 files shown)


jserv:
> But I have a question about later chapter.
> In chapter 0.7 introduces the /proc file system, use `proc_ops` structure in /proc is fine but that is defined in `linux/proc_fs.h` and you haven't explain it so I may confuse about what is that and the differences between `proc_ops` and `file_operations`.

Use `%` to remark the consideration with LaTeX script.

lkmpg.tex
```diff
@@ -649,30 +649,32 @@ \subsection{Device Drivers}
 
 \section{Character Device drivers}
 \label{sec:orgd3ad4b9}
-\subsection{The proc\_ops Structure}
+\subsection{The file\_operations Structure}
 \label{sec:org3950990}
-The proc\_ops structure is defined in \textbf{/usr/include/linux/fs.h}, and holds pointers to functions defined by the driver that perform various operations on the device. Each field of the structure corresponds to the address of some function defined by the driver to handle a requested operation.
+The file\_operations structure is defined in \textbf{/usr/include/linux/fs.h}, and holds pointers to functions defined by the driver that perform various operations on the device. Each field of the structure corresponds to the address of some function defined by the driver to handle a requested operation.
```
jserv, line 654: Use `\verb|file_operations|` instead.

jserv: SInce Linux v5.6, the interface has been replaced with `proc_ops`. You should denote.

lkmpg.tex
```diff
@@ -681,40 +683,49 @@ \subsection{The proc\_ops Structure}
...
     ssize_t (*splice_read)(struct file *, loff_t *, struct pipe_inode_info *, size_t, unsigned int);
-    int (*setlease)(struct file *, long, struct file_lock **);
+    int (*setlease)(struct file *, long, struct file_lock **, void **);
     long (*fallocate)(struct file *file, int mode, loff_t offset,
-	      loff_t len);
-    int (*show_fdinfo)(struct seq_file *m, struct file *f);
-};
+        loff_t len);
+    void (*show_fdinfo)(struct seq_file *m, struct file *f);
+#ifndef CONFIG_MMU
```
jserv, line 690: Remove this `#ifndef`-`#endif` block since we want to minimize code listing.

lkmpg.tex
```diff
@@ -681,40 +683,49 @@ \subsection{The proc\_ops Structure}
...
+} __randomize_layout;
 \end{code}
 
-Some operations are not implemented by a driver. For example, a driver that handles a video card won't need to read from a directory structure. The corresponding entries in the proc\_ops structure should be set to NULL.
+Some operations are not implemented by a driver. For example, a driver that handles a video card won't need to read from a directory structure. The corresponding entries in the file\_operations structure should be set to NULL.
 
 There is a gcc extension that makes assigning to this structure more convenient. You'll see it in modern drivers, and may catch you by surprise. This is what the new way of assigning to the structure looks like:
 
 \begin{code}
-struct proc_ops fops = {
```
jserv, line 696: Check `examples/procfs1.c` and you shall explain the interface change for Linux v5.6+.

asas1asas200: But the `proc_fs` structure is only for /proc file system and this chapter is introducing about character device drivers, should we explain it in next chapter?

lkmpg.tex
asas1asas200: Sorry, do I misunderstand it?
I think `proc_ops` is for files in the VFS, and it is a reduced structure from `file_operations` so normally we still use `file_operations`.

jserv: Reference [proc: faster open/read/close with "permanent" files](https://lwn.net/Articles/812612/) and quote the key considerations.

lkmpg.tex
jserv: Then, this patch should touch back and forth.

jserv (changes requested):
Please rebase the latest master branch to resolve the conflicts.

asas1asas200:
I have denoted the API transition of `file_operations` and tried to add some introductions of `proc_ops`, please review it and give me some comments if possible.
Thanks a lot.

lkmpg.tex
```diff
@@ -793,17 +801,19 @@ \subsection{The proc\_ops Structure}
...
+	.release = device_release
 };
 \end{code}
 
 The meaning is clear, and you should be aware that any member of the structure which you do not explicitly assign will be initialized to NULL by gcc.
 
-An instance of struct proc\_ops containing pointers to functions that are used to implement read, write, open, \ldots{} syscalls is commonly named fops.
+An instance of struct \verb|file_operations| containing pointers to functions that are used to implement read, write, open, \ldots{} syscalls is commonly named fops.
+
+In Linux v5.6+, this structure has been instead by \ref{sec:proc_ops} in the \textbf{/proc} file system.
```
jserv, line 816: Change to the following:
> Since Linux v5.6, the `proc_ops` structure was introduced to replace the use of of the `file_operations` structure when registering proc handlers.

lkmpg.tex
```diff
@@ -941,6 +951,13 @@ \section{The /proc File System}
 
 \samplec{examples/procfs1.c}
 
+\subsection{The proc\_ops Structure}
+\label{sec:proc_ops}
+The \verb|proc_ops| structure is defined in \textbf{/usr/include/linux/proc\_fs.h} in Linux v5.6+.
+In old kernel, it used \verb|file_operations| for custom hooks in \textbf{/proc} file system, but it contains some members that are unnecessary in VFS, and every time VFS expands \verb|file_operations| set, \textbf{/proc} code comes bloated.
```
jserv, line 957: Should be "In older kernels, ..."

jserv:
This pull request was supposed to close #6

jserv:
Thank @alan23273850 for contributing! I amended the commit messages. Never use markdown syntax within the commit messages and program comments because they are intended to be plain text.

PR #15, opened 2021-07-28
feat: Generate PDF/HTML on release event with GitHub Actions

et84121:
this feature will trigger github action to generating book on [github released event](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#release) and the book will be uploaded to release assets

## Github Page Setting
repo owner should set repo setting like this pic
image

PR #17, opened 2021-07-28
Add Pandoc convert GitHub Page script

focaaby:


Makefile
```diff
@@ -1,6 +1,9 @@
 all: lkmpg.tex
 	pdflatex -shell-escap lkmpg.tex
 
+html:
+	pandoc -s lkmpg.tex -o lkmpg.html
```
jserv, line 5: I checked the HTML file generated by `pandoc`, and It appears that HTML file is not recognizable for source code listing.

Related: https://github.com/sysprog21/lkmpg/pull/15

focaaby: It seems that `pandoc` did not include the `lib/codeblock.tex` properly.

jserv: Since `make4ht` renders HTML better than `pandoc`, I would like to wait for #20 prior to this pull request.

jserv:
Drop pandoc support in favor of #20

PR #23, opened 2021-07-31
Update comment to match this example

demonsome:
The name of the device file created in this example is "chardev". 
So  a input command "sudo cat /dev/chardev" to the terminal will surely return a valid message of "I already told you %d times Hello world!"
I'm not sure why author indicates us to check "mycharfile", which wasn't even mentioned in the book.


jserv:
Thank @demonsome for contributing!

PR #20, opened 2021-07-30
Generate HTML via make4ht

fennecJ:
`make4ht` has a lot of defectors:
* When it parsing `lkmpg.tex` and meet `\tableofcontents` will throw error
* When it parsing `lkmpg.tex` and meet url with underscore `\_`, it will output `\_` 
directly instead of `_`
* TOC in generated page isn't as pretty as pdf version
* The `github action` script released by make4ht's team
worked inconsistent  with computer side which causing
error when generating `github page`,and I currently cannot 
find a way to generate `github
page` through `github action` via `make4ht`.
You can see the result page in [here](https://fennecj.github.io/lkmpg/html) currently.

Issue #27, opened 2021-08-01
Reproducible TeX build environment using Docker


ekangmonyet:
This is to centralize efforts of developing Github Actions and **_if needed_** a customized docker container for the purpose of building all assets (PDFs, HTMLs, etc).

- [x] PDF Building (#28)
- [x] Existing release fix (#29 ~~or https://github.com/softprops/action-gh-release/pull/134~~)
- [x] HTML Building (#32 , ~~#20~~)

ekangmonyet:
I think there's a cyclic dependency between this and #20 , my opinion is to merge #20 first, start this and improve HTML building at the same time.

Hsins:
Just focus on "building docker image for consistent environment" and then the issue could be the first task to be done. The title of this issue are almost the same with #25.

ekangmonyet:
I will focus on producing the PDF file first instead, since it is pretty stable and no other task dependencies. It will serve as a foundation to build other assets too.

fennecJ:
For html building, the following script worked for me:
``` yml
name: Generate HTML and deploy to GitHub Page.

on:
  push:
    branches:
      - master

jobs:
  build-deploy:
    runs-on: ubuntu-latest
    steps:
      - name: checkout branch
        uses: actions/checkout@v2

      - name: install dependencies and generating files
        uses: xu-cheng/texlive-action/full@v1
        with:
          run: |
            apk add make
            make html

      - name: deploy site
        uses: peaceiris/actions-gh-pages@v3
        with:
          github_token: ${{ secrets.GITHUB_TOKEN }}
          publish_dir: ./html                                          # --> should be the same as your output folder of make4ht
          publish_branch: gh-pages
```
And you can checkout the output in [here](https://fennecj.github.io/lkmpg/)
Looks like the page rendered well.

ekangmonyet:
#32 is working on it.

Hsins:
@fennecJ @ekangmonyet 

Currently both workflow [texlive-action](https://github.com/xu-cheng/texlive-action) and container [texlive](https://hub.docker.com/r/texlive/texlive) are build for general use, and they're not actually an "reproducible TeX build environment" for our project.

I've build [this docker image](https://github.com/TeXtw/docker-lkmpg) supports full TeXLive with Gnuplot and Pandoc. If you have installed docker on your local machine, try with following steps to compile with it:

```bash
# clone the source code
$ git clone https://github.com/sysprog21/lkmpg && cd lkmpg

[...]

PR #32, opened 2021-08-01
Publish GitHub Page with GitHub Action

focaaby:
Use #20 `makefile` 
1. Generate the HTML assets and deploy them to the [GitHub Page](https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site)[1].
    * Here is the demo URL: https://focaaby.github.io/lkmpg/index.html
2. Publish the `html` directory as release HTML assets.
    * Demo download URL: https://github.com/focaaby/lkmpg/releases/download/latest/lkmpg-html.tar.gz

References:
1. https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site

### DIFF
--- a/.github/workflows/deploy_github_page.yaml
+++ b/.github/workflows/deploy_github_page.yaml
@@ -1,0 +1,24 @@
+name: build-deploy-github-page
+
+on:
+  push:
+    branches: [ master ]
+
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    container: twtug/lkmpg
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Build
+        run: |
+          make html
+      - name: Deploy to gh-pages branch
+        uses: peaceiris/actions-gh-pages@v3
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          publish_dir: ./html
+          publish_branch: gh-pages
--- a/.github/workflows/generate_doc.yml
+++ b/.github/workflows/generate_doc.yml
@@ -9,22 +9,29 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
-    container: texlive/texlive
+    container: twtug/lkmpg
 
     steps:
       - uses: actions/checkout@v2
       - name: Build
-        run: make all
+        run: |
+          make all
+          make html
+          tar zcvf lkmpg-html.tar.gz ./html
       - name: Delete old release asset
         uses: mknejp/delete-release-assets@v1
         with:
           token: ${{ github.token }}
           fail-if-no-assets: false
           tag: latest
-          assets: lkmpg.pdf
+          assets: |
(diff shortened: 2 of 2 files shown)


jserv:
> Here is the demo URL: https://focaaby.github.io/lkmpg/index.html

TikZ graph was not rendered correctly. @fennecJ, can you give a hint?

fennecJ:
I tryied to use @focaaby 's action script, and my TikZ graph was not rendered correctly, either.
I have checked the version of the tool we used and don't find any problem currently.

As I mentioned in #27 , [xu-cheng/texlive-action](https://github.com/xu-cheng/texlive-action) worked well and rendered the page correctly.
Consider switch to [xu-cheng/texlive-action](https://github.com/xu-cheng/texlive-action) instead of texlive/texlive ?

Hsins:
Umm...

- Workflows in GitHub Actions/GitLab CI/Azure Pipeline... are series of tasks running on the docker container.
- Docker containers are instances of given docker image.

They are not the same thing. Switching from workflows to workflows is just switching the environments that other defined, why don't we just created the environment for us?

fennecJ:
@Hsins 
Sorry I didn't notice that you have built up a docker image when I comment then.
And I tried your [docker image](https://github.com/TeXtw/docker-lkmpg), it worked well.
I agree that we should just create the environment for us.

jserv (changes requested):
Read this carefully: https://chris.beams.io/posts/git-commit/
* Wrap the body at 72 characters

Don't use markdown syntax inside git commit message.

jserv:
For the sake of naming consistency, rename `.github/workflows/generate_pdf.yml` to `.github/workflows/generate_docs.yml` which generates PDF and HTML accordinly.

jserv:
The message is too long:
> Add a new Github Action to publish to the GitHub page. Use https://github.com/TeXtw/docker-lkmpg image for reproducibility.

Wrap the body at 72 characters.

jserv:
Thank @focaaby for contributing!

jserv:
I was not aware of the HTML generation on GitHub Pages. Here are the configurations:
GitHub Pages

@focaaby, how should I make changes?

fennecJ:
For the sake of naming consistency, both 2  script files should be ended with `.yml` either `.yaml` .
According to [yaml.org](https://yaml.org/faq.html)
>### Is there an official extension for YAML files?
>Please use ".yaml" when possible.

It seems .yaml is better choise.

fennecJ:
@jserv 
I can see the page in [https://sysprog21.github.io/lkmpg](https://sysprog21.github.io/lkmpg) now.
Maybe it's caused by cache?
Try using ctrl + F5 to refresh the page.

jserv:
> I can see the page in https://sysprog21.github.io/lkmpg now.
> Maybe it's caused by cache?

It works for me now. Thanks!

PR #33, opened 2021-08-02
Making the file name "/proc/helloworld" same as others

demonsome:
In this paragraph,  the file name "proc helloworld" in the second sentance is not consistent with that of the latter sentences. The latter sentences use the file name "/proc/helloworld".
Otherwise, I suggest that the  "the file proc helloworld" in the second sentance should change to "the file proc named helloworld".


jserv:
Thank @demonsome for contributing!

PR #38, opened 2021-08-04
Correspond the example output of hello-5.c

RinHizakura:
For the example module `hello_5`, the book shows the incorrect output which does not correspond to the example input. They are revised here in case to confuse the reader.

Note that I also change `myintArray[2] = {-1, -1}` to `myintArray[2] = {420, 420}`. I think it could make reader clearly reliaze how the command line arguments change the output from the example input.

### DIFF
--- a/examples/hello-5.c
+++ b/examples/hello-5.c
@@ -13,7 +13,7 @@ static short int myshort = 1;
 static int myint = 420;
 static long int mylong = 9999;
 static char *mystring = "blah";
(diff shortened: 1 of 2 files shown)


jserv:
Thank @RinHizakura for contributing! The git commit messages should be improved through.

RinHizakura:
I'm sorry about that. I'll take care of it for next time's PR.

Issue #36, opened 2021-08-02
Text does not scale on firefox


nowakf:
Book text doesn't scale in Firefox 90.0.2, also chromium. It makes it pretty hard to read if you have bad eyesight.

ekangmonyet:
I would advocate for [Firefox Reader View](https://support.mozilla.org/en-US/kb/firefox-reader-view-clutter-free-web-pages).  If the reflow is accurate and stable, it can also be released as epub format.

fennecJ:
I think the reason why text can't be scale is because in `html.cfg`, line 13,
```css
font-size: 1vw;
```
When an element using `vw` as size unit, its size will be depended on the
width of the browser window size and won't be effected by scaling ratio
of browser. I tried to use `em` or `px` instead, however when scaling the
text, the code blocks' layout will be malformed and I currently can't find
a way to fix that.

jserv:
Close via commit a411938

PR #43, opened 2021-08-05
Rewrite CSS to support scaling

dancek:
This changes the layout to use the default font size with a reasonable
max width and margins around the body. Elements that would overflow on
narrow displays are changed to wrap (.author) or present a scroll bar as
necessary (figure, .verbatim, .fancyvrb). The properties on html and
body are set such that there is no full-document vertical scrolling.
Inline math is displayed in the same size as the surrounding text.

### DIFF
--- a/html.cfg
+++ b/html.cfg
@@ -2,15 +2,31 @@
 
 \Configure{tableofcontents*}{chapter,section,subsection}
 
-\Css{* :not(img) {
-    max-width: 100\%;
-    width: 50vw; 
-    height: auto;
-    margin: 0 auto;
+\Css{html {
+    width: 100vw;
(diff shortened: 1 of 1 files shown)


dancek:
This fixes #36.

While writing this I noticed that some files get their line numbers as `tspan` instead of `span.ecrm-0500`. That's very strange and sounds like tex4ht is misbehaving (tspan is an svg element and shouldn't be present in html).

fennecJ (approved):
I've tried your script and it worked will on my machine.
Also there are no `tspan` elements in generated html files.

jserv:
Thank @dancek for contributing!

PR #50, opened 2021-08-09
fix minor typo

VxTeemo:
verb warp wrong at line 330 "linux/.config"


jserv:
Thank @VxTeemo for contributing!

PR #51, opened 2021-08-09
Fix init message

ChinYikMing:
I think the filename should be /proc/sleep


jserv:
Thank @ChinYikMing for contributing!

PR #62, opened 2021-08-16
Provide detailed info for reference count.

fennecJ:
This patch is intented to provide detailed info for reference count.

I add sudo lsmod command to provide an alternative way for user to check
the reference count for specific module.

I think `reference count` would be better compared to `use count` so I
replace `use count` with `reference count`.

The paragraph says "there are functions which let you increase,
decrease and display this counter".
But it didn't provide the function to display the the reference counter.
So I add description for module_refcount(THIS_MODULE) as the function
to display the the reference counter.

### DIFF
--- a/lkmpg.tex
+++ b/lkmpg.tex
@@ -894,14 +894,15 @@ The results of this would be impossible to predict, but they can not be very pos
 Normally, when you do not want to allow something, you return an error code (a negative number) from the function which is supposed to do it.
 With \cpp|cleanup_module| that's impossible because it is a void function.
 However, there is a counter which keeps track of how many processes are using your module.
-You can see what its value is by looking at the 3rd field of \verb|/proc/modules|.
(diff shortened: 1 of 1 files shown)


lkmpg.tex
```diff
@@ -894,14 +894,15 @@ \subsection{Unregistering A Device}
 Normally, when you do not want to allow something, you return an error code (a negative number) from the function which is supposed to do it.
 With \cpp|cleanup_module| that's impossible because it is a void function.
 However, there is a counter which keeps track of how many processes are using your module.
-You can see what its value is by looking at the 3rd field of \verb|/proc/modules|.
+You can see what its value is by looking at the 3rd field with \sh|cat /proc/modules| or \sh|sudo lsmod|.
```
jserv, line 897: Change "with" to "with the command"

fennecJ: Done

jserv:
Thank @fennecJ again for contributing. I amended the git commit message. Please check the seven rules: https://chris.beams.io/posts/git-commit/
> Do not end the subject line with a period

Issue #64, opened 2021-08-16
Mention PDF download information in generated HTML


jserv:
At present, HTML is rendered via `make4ht`. However, it lacks the information about PDF downloading.

The expected output:
1. Change the HTML template and/or setting to add the PDF download hyperlink.
2. Mention PDF download only in generated HTML. The body of PDF document should not contain the download link though.

fennecJ:
I come up with 2 solutions for this issue
* 1 We can add
  ```
  \begin{center}
  \href{https://github.com/sysprog21/lkmpg/releases/download/latest/lkmpg.pdf}{Download this page in pdf}
  \end{center}
  ```
  At the end of `html.cfg`, and the information about PDF downloading can only be at the top of the page:
  ![image](https://user-images.githubusercontent.com/58484289/129660772-69928f4b-045b-4cdd-956b-bc4c9295b296.png)

* 2 Using \ifdefined\HCode
  By this way we can choose where to place the information about PDF downloading.
  ![image](https://user-images.githubusercontent.com/58484289/129661405-2cd77704-22f8-46e1-9b78-b697090acd8d.png)

jserv:
How about following [Yocto Project Overview and Concepts Manual](http://docs.yoctoproject.org/overview-manual/)? We can put the hyperlinks to PDF download and GitHub  in the upper right corner of the HTML page.

fennecJ:
Add following code
```diff
--- a/html.cfg
+++ b/html.cfg
@@ -45,5 +45,15 @@
     user-select: none;
 }}
 
+\Css{.flushright:first-child {
+    position:absolute;
+    top:0;
+    right:0;
+}}
+
 \begin{document}
 \EndPreamble
+
+\begin{flushright}
+\href{https://github.com/sysprog21/lkmpg/releases/download/latest/lkmpg.pdf}{Download this page in pdf}
+\end{flushright}
```
into `html.cfg` would be looked like:
![image](https://user-images.githubusercontent.com/58484289/129674515-6d2fd051-fc31-4c3b-83a4-78eb58bf5232.png)

You can check it in [https://fennecj.github.io/lkmpg/](https://fennecj.github.io/lkmpg/)

jserv:
Change from "Download this page in pdf" to "Download PDF document"

jserv:
> You can check it in https://fennecj.github.io/lkmpg/

Can you reserve some spaces at the end of the line "Download PDF document"?

fennecJ:
[...]

PR #67, opened 2021-08-17
Add pdf download link in html only

fennecJ:
You can check it in [https://fennecj.github.io/lkmpg/](https://fennecj.github.io/lkmpg/)

### DIFF
--- a/html.cfg
+++ b/html.cfg
@@ -45,5 +45,15 @@
     user-select: none;
 }}
 
+\Css{.flushright:first-child {
+    position:absolute;
+    top: 10px;
+    right: 50px;
+}}
+
 \begin{document}
 \EndPreamble
+
+\begin{flushright}
+\href{https://github.com/sysprog21/lkmpg/releases/download/latest/lkmpg.pdf}{Download PDF document}
+\end{flushright}


jserv:
Append `Close #64` to git commit message, so that it can be automatically close once this pull request is merged.

html.cfg
```diff
@@ -45,5 +45,15 @@
...
+    position:absolute;
+    top: 10px;
+    right: 50px;
+}}
+
 \begin{document}
 \EndPreamble
+
+\begin{flushright}
+\href{https://github.com/sysprog21/lkmpg/releases/download/latest/lkmpg.pdf}{Download pdf document}
```
jserv, line 58: Use uppercase for PDF. That is, `PDF`.

fennecJ:
> Append `Close #64` to git commit message, so that it can be automatically close once this pull request is merged.

Cool, I didn’t know that before.

jserv:
> Cool, I didn’t know that before.
> Thanks for informing me about that.

See https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

PR #69, opened 2021-08-21
Update some file path to kernel v5.x in comment of example/kbleds.c

gagachang:
Update below file path in comment of example/kbleds.c to kernel version v5.x:
1. drivers/char/vt_ioctl.c -> drivers/tty/vt/vt_ioctl.c
2. drivers/char/keyboard.c -> drivers/tty/vt/keyboard.c


jserv:
Thank @gagachang for contributing!

PR #75, opened 2021-08-24
Fix: errno assignment should be comparison.

tuckerpo:


jserv:
Thank @tuckerpo for contributing!

PR #85, opened 2021-08-31
Add a Github button in html files

### LINKED ISSUE #84: Show GitHub button in rendered HTML
At present, the readers can only reach GitHub hyperlink at the end of browsing the rendered HTML. It would be great if we can add the GitHub button in the top right side of HTML, so that someone can check this repository and/or promote via further actions.

fennecJ:
The patch using a CDN script to render a pretty Github button.
Close #84

### DIFF
--- a/html.cfg
+++ b/html.cfg
@@ -51,9 +51,20 @@
     right: 50px;
 }}
 
+\Css{.right {
+    text-align: right;
+}}
+
+\AtBeginDocument{%
+\Configure{@HEAD}{\HCode{
+    <script async defer src="https://buttons.github.io/buttons.js"></script>
+    <div class="right">
+        <a class="github-button" href="https://github.com/sysprog21/lkmpg" data-size="large" aria-label="View on GitHub">View on GitHub</a>
+        <a class="github-button" href="https://github.com/sysprog21/lkmpg/releases/download/latest/lkmpg.pdf" data-icon="octicon-download" data-size="large" aria-label="Download PDF document">Download PDF document</a>
+        <br><br>
+    </div>
+    \Hnewline}}
+}
+
 \begin{document}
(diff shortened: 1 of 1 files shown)


fennecJ:
Output will be looked like following:
image

html.cfg
```diff
@@ -51,9 +51,21 @@
...
 
-\begin{flushright}
-\href{https://github.com/sysprog21/lkmpg/releases/download/latest/lkmpg.pdf}{Download PDF document}
-\end{flushright}
\ No newline at end of file
+\AtBeginDocument{%
+\Configure{@HEAD}{\HCode{
+    <script async defer src="https://buttons.github.io/buttons.js"></script>
+    <div class="right">
+        <a class="github-button" href="https://github.com/sysprog21/lkmpg" data-size="large" aria-label="Check the repo on GitHub">Check the repo on GitHub</a>
```
jserv, line 62: Change "Check the repo on GitHub" to "View on GitHub"

html.cfg
```diff
@@ -51,9 +51,21 @@
...
+        <br><br>
+        <a href="https://github.com/sysprog21/lkmpg/releases/download/latest/lkmpg.pdf">Download PDF document</a>
```
jserv, line 64: Is it possible to use the similar button to download PDF? https://buttons.github.io/

fennecJ: Yes, however it cause different width between 2 buttons:
image
And currently author don't provide a way to resize the button.[1]

To align them, we need to assure they have same number of characters.
Or maybe put them in the same line:
image

[1] - https://github.com/ntkme/github-buttons/issues/34

jserv: I think it makes sense to put two buttons on the same line.

PR #88, opened 2021-09-02
Follow the Linux coding style

linD026:
Except the indention is four spaces rather than tabs
for sticking to compact layout of source listing,
we should follow others rules.

The Linux coding style
- https://www.kernel.org/doc/html/latest/process/coding-style.html

### DIFF
--- a/examples/Makefile
+++ b/examples/Makefile
@@ -36,3 +36,6 @@ all:
 clean:
 	$(MAKE) -C /lib/modules/$(shell uname -r)/build M=$(PWD) clean
 	$(RM) other/cat_noblock *.plist
+
+indent:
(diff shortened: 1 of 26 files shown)


examples/Makefile
```diff
@@ -36,3 +36,6 @@ all:
 clean:
 	$(MAKE) -C /lib/modules/$(shell uname -r)/build M=$(PWD) clean
 	$(RM) other/cat_noblock *.plist
+
+clang-format:
```
jserv, line 40: Don't do that. There is a "indent" target in top-level `Makefile`.

jserv:
Thank @linD026 for the efforts!

PR #91, opened 2021-09-03
Fix Typo

ccs100203:
from "mask" to "masks"
from "quick" to "quickly"
from "a" to "an"


jserv:
Thank @ccs100203 for contributing!

Issue #72, opened 2021-08-23
Set up CI infrastructure to automate Linux kernel module validation


jserv:
At present, CI only checks the coding style. It is far from complete validation. We shall set up the CI infrastructure which is able to perform:
1. Build all example Linux modules within Ubuntu Linux 20.04 (kernel 5.4+);
2. Attempt to do "insmod" and "rmmod";
3. (Optionally) check the kernel messages and validate;

See https://github.com/kernelci/kernelci-core as an example.

fennecJ:
I wrote a simple shell script trying "insmod" and "rmmod" all module on my machine and meet following errors:
`chardev2.ko`:
At first time insmod and rmmod didn't throw any errors, however when using command `dmesg`, it shows:
```bash
[   78.479322] chardev2: loading out-of-tree module taints kernel.
[   78.479463] chardev2: module verification failed: signature and/or required key missing - tainting kernel
[   78.480170] Device created on /dev/char_dev
```
And when I tried to insmod the module again, terminal shows:
```bash
insmod: ERROR: could not insert module chardev2.ko: Device or resource busy
```
And the output of `dmesg` was:
```bash
[  157.168167] Sorry, registering the character device  failed with -16
```
---
`bottomhalf.ko`:
`intrpt.ko`:
When insmod `bottomhalf.ko` or `intrpt.ko`, the error was something like:
```bash
insmod: ERROR: could not insert module xxx.ko: Unknown error 517
```
And the output of `dmesg` was:
```bash
Unable to request GPIOs for LEDs: -517
```
I think my board's GPIO number may be different from sample code, which cause this error.
But I currently don't know how to check GPIO number of my board.

---
`syscall.ko`
When insmod `syscall.ko` first time, the terminal simply shows
```bash
killed
```
And the output of `dmesg` was:
```bash
[  384.864739] BUG: unable to handle page fault for address: ffff8aad47458000
[  384.864755] #PF: supervisor read access in kernel mode
[  384.864757] #PF: error_code(0x0000) - not-present page
[  384.864758] PGD 1aa01067 P4D 1aa01067 PUD 4907ff067 PMD 4907fe067 PTE 0
[  384.864761] Oops: 0000 [#1] SMP NOPTI
[  384.864763] CPU: 3 PID: 3017 Comm: insmod Tainted: G           OE     5.11.0-27-generic #29~20.04.1-Ubuntu
[...]

PR #95, opened 2021-09-06
Validate examples in CI pipeline

fennecJ:


fennecJ:
Now there are 4 modules in `NotTest.txt`:
`bottomhalf.ko`
`intrpt.ko`
`kbleds.ko`
`start.ko`
For the first 2 modules, they are excluded because they are intended to run on raspberry pi board.
For `kbleds.ko`, it will be killed in docker and raise error.
For `start.ko`, as I mentioned in #72, now it has error and haven't be fixed yet. So I put it in `NotTest.txt` temporarily.

.ci/build-n-run.sh
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash
+uname -r
+cd examples
+make
+cd ..
+exclude_list=$(cat .ci/NotTest.txt);   #the module in NotTest.txt will be ignored by the shell
```
jserv, line 6: `NotTest` is ambiguous. It is hard to distinguish from "not tested" and "non-working." In this case, you meant "non-working" modules which should be excluded.

.ci/build-n-run.sh
```diff
@@ -0,0 +1,48 @@
...
+        fi
+    done
+    if [ "$exclude" = true ]; then 
+        echo "$filename" "is excluded!"; continue;
+    else
+        echo "$filename"
+        sudo insmod "$filename"
+        if [ $? = 0 ]; 
+        then 
+            sudo rmmod "$filename"
```
jserv, line 22: You can simply rewrite as:
```shell
sudo rmmod "$filename" || exit 1
```

jserv:
Use `git rebase -i` to rework/squash the commits for reviewing purpose.

.ci/build-n-run.sh
```diff
@@ -0,0 +1,26 @@
...
+    exclude=false
+    for file in $exclude_list; do
+        if [ "$filename" = "$file" ]; then
+            exclude=true
+        fi
+    done
+    
+    if [ "$exclude" = true ]; then
+        echo "$filename" "is excluded!"; continue;
+    else
```
jserv, line 20: Remove unnecessary `else` statement.

.ci/build-n-run.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+uname -r
+#build
+cd examples
```
jserv, line 4: You don't have to run `cd`. Instead, use `make -C examples`.

.ci/build-n-run.sh
```diff
@@ -0,0 +1,26 @@
...
+make
+cd ../
+#run module test
+exclude_list=$(cat .ci/non-working.txt);   #the module in non-working.txt will be ignored by the shell
+for filename in examples/*.ko; do
+    [ -e "$filename" ] || continue
+    exclude=false
```
jserv, line 11: Simplify the handling. You can avoid the use of variable `exclude`.

.ci/build-n-run.sh
```diff
@@ -0,0 +1,26 @@
...
+            exclude=true
+        fi
+    done
+    
+    if [ "$exclude" = true ]; then
+        echo "$filename" "is excluded!"; continue;
+    else
+        echo "$filename"
+#test 2 times
+        ( sudo insmod "$filename" && sudo rmmod "$filename" ) || exit 1;
```
jserv, line 23: You should trim `$filename` to remove the `.ko` suffix for `rmmod`.

jserv (changes requested):
Append "Close #72" at the end of git commit message.

.ci/build-n-run.sh
jserv: You have to check the return value of `make` respectively, otherwise CI has no further actions.

.ci/build-n-run.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
```
jserv, line 1: Change the [shebang](https://en.wikipedia.org/wiki/Shebang_(Unix)) to `#!/usr/bin/env bash`, which allows the overriding.

.ci/build-n-run.sh
jserv: You can make use of bash function to modularize the routines.
E.g., https://github.com/sysprog21/kvm-host/blob/master/scripts/build-linux-image.sh

fennecJ:
The script worked well on my machine, however it meet problem on Github action.
I am still inspecting why it work failed.

.ci/build-n-run.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
+
+function build_example_code()
```
jserv, line 3: Change to `build_examples`.

.ci/build-n-run.sh
```diff
@@ -0,0 +1,29 @@
...
+{   
+    make -C examples/ || exit 1;
```
jserv, line 5: Shorten as `make -C examples` and tailing `;` is not required.

.ci/build-n-run.sh
```diff
@@ -0,0 +1,29 @@
...
+function gen_test_file_list()
+{
+    exclude_list=$(cat .ci/non-working.txt);                #the module in non-working.txt will be ignored by the shell
+    file_list=$(ls examples/*.ko | xargs -n 1 basename);    #only the base name will be shown
+
+    #wipe out file existed both in exclude_list and file_list
+    to_be_test=$(echo "${file_list[@]}" "${exclude_list[@]}" | tr ' ' '\n' | sort | uniq -u)
+}
+
+function test_example_code()                                #test example code 2 times
```
jserv, line 17: Change to `run_examples`.

.ci/non-working.txt
```diff
@@ -0,0 +1,4 @@
+bottomhalf.ko
```
jserv, line 1: The suffix `.ko` is redundant. Instead, you can add back in shell scripts.

.ci/build-n-run.sh
```diff
@@ -0,0 +1,29 @@
...
+{
+    for filename in $to_be_test; do
```
jserv, line 19: Change `filename` to `mod` or `module`.

.ci/build-n-run.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
+
+function build_example_code()
+{   
+    make -C examples/ || exit 1;
+}
+
+function gen_test_file_list()
```
jserv, line 8: Alternatively, you can create a function such as `run_mod`, which accepts one parameter as the specified Linux kernel module. Then, use that function here.

fennecJ: Sorry, I don't understand your meaning very well.
So I guess `run_mod` is supposed to run kernel module with accepted parameter only. (Without any string modification)
And I don't know why put it in `gen_test_file_list()`?
Can you give further explanations?

linD026:
The `start.c` and `stop.c` are the same module.
Did you try this command?
```shell
obj-m += startstop.o 
startstop-objs := start.o stop.o 
```

fennecJ:
@linD026 It works, thank you.

.ci/build-n-run.sh
jserv: It would better if you can implement another function which returns the filtered list of available Linux kernel modules, which can be then used with the `for` statement whose body consists of the invocation of `run_mod`.

That is,
```bash
for mod in $(mod_list)
do
    run_mod $mod
done
```

jserv: See https://www.hostinger.com/tutorials/bash-for-loop-guide-and-examples/

jserv: It is not elegant to rely on the function `gen_test_file_list` though. It does not make sense because you can filter out the non-working modules with the UNIX commands in combination.

jserv:
> According to [this commit](https://github.com/torvalds/linux/commit/686fef928bba6be13cabe639f154af7d72b63120#diff-27962f93d3af82317462213521ad9983f6cdbc2317f334d3ccb9501588669f48), the timer API has been changed.
> You can see the current version(5.14) at [here](https://elixir.bootlin.com/linux/latest/source/include/linux/timer.h#L131).
> lwn article: [Improving the kernel timers API](https://lwn.net/Articles/735887/).
> I alter some codes to make it work:
[...]

@linD026, create a new issue.

.ci/build-n-run.sh
```diff
@@ -0,0 +1,33 @@
...
+{
+    exclude_list=$(cat .ci/non-working.txt);                                  #module in non-working.txt will be ignored
+    file_list=$(ls examples/*.ko | xargs -n 1 basename | sed 's/\.ko$//');    #show the filename without extension and path
+
+    #wipe out file existed both in exclude_list and file_list
+    to_be_test=$(echo "${file_list[@]}" "${exclude_list[@]}" | tr ' ' '\n' | sort | uniq -u)
+}
+
+function run_examples()                                                       #test example code 2 times
```
jserv, line 22: Move the tested-twice logic to `run_mod`.

.ci/non-working.txt
```diff
@@ -0,0 +1,3 @@
+bottomhalf
```
jserv, line 1: The suffix `.txt` is redundant. In UNIX-like systems, the suffix does not help a lot.

.ci/build-n-run.sh
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
```
jserv, line 1: You should do `chmod +x .ci/build-n-run.sh`

.ci/build-n-run.sh
```diff
@@ -0,0 +1,30 @@
...
+
+function run_mod() #test module 2 times
+{
+    ( sudo insmod "examples/$1.ko" && sudo rmmod "$1" ) || exit 1;
+    ( sudo insmod "examples/$1.ko" && sudo rmmod "$1" ) || exit 1;
+}
+
+function run_examples()
+{
+    for module in $mod_list; do
```
jserv, line 22: Rewrite as following:
```bash
for module in $(get_mod_list); do
    run_mod "$module"
done
```

jserv: You can rename `get_mod_list` to `list_mod` for consistency.

.ci/build-n-run.sh
```diff
@@ -0,0 +1,30 @@
...
+        echo "$module"
+        run_mod "$module"
+    done
+}
+
+build_example
+get_mod_list
```
jserv, line 29: `get_mod_list` should be called in the body of `run_examples`.

.ci/build-n-run.sh
fennecJ: It rasing error when I add '('  before `list_mod`
```bash
.ci/build-n-run.sh: line 24: list_mod: command not found
```
My code:
```bash
function run_examples()
{
    get_mod_list
    for module in $(list_mod); do
        echo "$module"
        run_mod "$module"
    done
}
```

.ci/build-n-run.sh
fennecJ: I add it in the second line.

.ci/build-n-run.sh
jserv: You can call the function in bash. See https://ryanstutorials.net/bash-scripting-tutorial/bash-functions.php
You have to dump something in function `list_mod` though.

.ci/build-n-run.sh
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bash
+chmod +x .ci/build-n-run.sh
```
jserv, line 2: No! I meant, you should make this file executable, so that we can track in git repository.
Remove `chmod` here.

jserv: Run `ls -al .ci` for comparisons.

.ci/build-n-run.sh
fennecJ: Cool, I don't know that before.
Thanks for letting me know that.

.ci/build-n-run.sh
```diff
@@ -0,0 +1,30 @@
...
+}
+
+function list_mod()
+{
+    #following list will contain all filename which doesn't appear in non-working without path and file extension
+    echo `ls examples/*.ko | xargs -n 1 basename | sed 's/\.ko$//' && cat .ci/non-working` | tr ' ' '\n' | sort | uniq -u
+                            #cut path             #cut file extension #dispose modules listed in non-working
+}
+
+function run_mod() #test module 2 times
```
jserv, line 15: Move the comment before `function run_mode`.

.ci/build-n-run.sh
```diff
@@ -0,0 +1,30 @@
...
+
+function build_example()
+{   
+    make -C examples || exit 1
+}
+
+function list_mod()
+{
+    #following list will contain all filename which doesn't appear in non-working without path and file extension
+    echo `ls examples/*.ko | xargs -n 1 basename | sed 's/\.ko$//' && cat .ci/non-working` | tr ' ' '\n' | sort | uniq -u
```
jserv, line 11: Can you rewrite with awk? The commands might be shortened.

Gomez2003:
This serves to ????

fennecJ:
@Gomez2003 
Seems like you are confusing about receiving e-mail from GitHub notification system.
And I guess why you receive that is because you set yourself as watcher of this repository.

.ci/build-n-run.sh
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
+
+function build_example()
+{   
+    make -C examples || exit 1
+}
+
+function list_mod()
+{
+    #following list will contain all filename which doesn't appear in non-working without path and file extension
```
jserv, line 10: Check the grammar. Should be
> all file names which are not specified in file non-working.

In UNIX-like systems, there is no distinct concept about "file extensions," and you can find various files starting with dot (`.`) such as `.bashrc` and `.config`.

Issue #82, opened 2021-08-26
Improve readability


Kellegram:
Hi, first of all, great write-up. I do want to mention however that the theme of the site could be improved for readability. With popularity of dark mode, choosing dark colors could be the obvious first step, however it's not best for everyone of course. I recommend opening the website in reading mode via firefox or other browser, they offer different colors and formatting to make it more accessible for everyone, including those with reading problems. I did a quick few changes via inspector to see how it would look otherwise. The font is not variable so I couldn't set the text to be semi-bold (to make it look less wiry in certain situations). Using a more readable font like manrope could be great for accessibility. 

Here is what I got by using colors from my website as well as increasing font-size by 10% (to 1.1em) of parent as well as increasing header4 size by 20% (1.2em) as well as content width by 10rem: https://i.imgur.com/NBCK4XU.png

A change of font could improve this, here are some example using firefox reading mode: 
- Dark version: https://i.imgur.com/ZQMXLZX.png
- Sepia: https://i.imgur.com/5VGR3tJ.png
- Dark with more content width (still within decent eye travel): https://i.imgur.com/R4C0n9I.png

I would love to create a Pull Request to help, however I am not sure where is the source of css for this project, I am unfamiliar with the process used to build this page. I tried running a lighthouse Audit via web.dev/measure but it gives a network error on your page. I ran it via Edge devtools instead and while the score is overall pretty good, including for accessibility, it does fail somewhat on contrast ratios. 

Themes are mostly subjective of course but it seems like we are moving into dark mode era, so perhaps it's something to consider, but at the very least using some kind of sepia colors or a theme like Solarized could vastly improve the experience for folk with very bright screens, people working at night, people with sensitive eyes, etc. Something in-between light and dark mode is always a good balance assuming good contrast ratios are maintained. 

The page is almost missing meta description which in turn would give it a worse pageRank score and will fail to display embeds that can summarise the content of the page. This is mostly an SEO issue but worth noting.

jserv:
After #96, the text with hyperlinks would look slightly larger than plain text, at least in Chrome web browser. Is there any effective tweak to make all text more consistent?

Kellegram:
> 
> 
> After #96, the text with hyperlinks would look slightly larger than plain text, at least in Chrome web browser. Is there any effective tweak to make all text more consistent?

[...]

PR #96, opened 2021-09-08
Improve readability

Kellegram:
Related issue #82 

This new foreground + background color maintains the minimum of 7:1 contrast ratio as specified by WCAG (this has 7.36:1) and is the same colors firefox reading mode uses when choosing the sepia theme. 

Fonts are too small by default, using responsive units (em) let's me increase them, I am unsure where or if there is font-size specified for parent anywhere but it seems to have been manually decreased.

Increased max-width slightly to adjust for new size and more standard line-height, so the doc doesn't feel significantly longer, still maintains good eye travel.

### DIFF
--- a/html.cfg
+++ b/html.cfg
@@ -8,11 +8,26 @@
 }}
 
 \Css{body {
-    max-width: 50rem;
+    max-width: 55rem;
     box-sizing: border-box;
     padding: 1rem;
     margin: 0 auto;
     overflow-x: hidden;
+    background-color: #F4ECD8;
+    color: #5B464B;
+    line-height: 1.5;
+}}
+
+\Css{a {
+color: #0060DF;
+}}
+
+\Css{p, a {
+font-size: 1.2em;
+}}
+
+\Css{p + pre {
+font-size: 1.1em;
 }}
 
 \Css{div.author {
--- a/html.cfg
+++ b/html.cfg
@@ -13,13 +13,13 @@
     padding: 1rem;
     margin: 0 auto;
     overflow-x: hidden;
-    background-color: #F4ECD8;
-    color: #5B464B;
+    background-color: \#F4ECD8;
+    color: \#5B464B;
     line-height: 1.5;
 }}
 
 \Css{a {
-color: #0060DF;
+color: \#0060DF;
 }}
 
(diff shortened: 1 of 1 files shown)


Kellegram:
Let me know if I edited the cfg correctly. Code syntax highlight could use a change if this were to be used. It, would need to have slightly darker colors, it's already close to having good contrast ratios for the most part.

Kellegram:
Also consider using better font, like Manrope. Manrope has a variable weight version and from my experience, people with accessibility issues don't struggle with it. Variable weight means you can use semi-bold which is great, it has pleasant shape and scales really well. (https://manropefont.com/)

jserv:
Manrope is an open-source modern sans-serif font family. These fonts are licensed under the Open Font License. It is great to LKMPG development.

html.cfg
```diff
@@ -8,11 +8,26 @@
 }}
 
 \Css{body {
-    max-width: 50rem;
+    max-width: 55rem;
     box-sizing: border-box;
     padding: 1rem;
     margin: 0 auto;
     overflow-x: hidden;
+    background-color: #F4ECD8;
```
fennecJ, line 16: Add '\\' before character '#', otherwise `make4ht` will see it as escape character and raise error.

html.cfg
```diff
@@ -8,11 +8,26 @@
...
+    color: #5B464B;
```
fennecJ, line 17: Ditto, add '\\' before character '#'.

html.cfg
```diff
@@ -8,11 +8,26 @@
...
+    line-height: 1.5;
+}}
+
+\Css{a {
+color: #0060DF;
```
fennecJ, line 22: Ditto, add '\\' before character '#'.

fennecJ (changes requested):
The output looks promising to me after fix those minor problems.

Kellegram:
Fixed it. if there is anything else that you want me to fix or change let me know. I find this version readable but it's good to have second opinion, perhaps you noticed something that doesn't look the way it was intended?

fennecJ (approved):
Looks good to me.

jserv:
Thank @Kellegram for contributing!

PR #102, opened 2021-09-12
fix typo: `compatibiity -> compatibility`

giladreti:


jserv:
Thank @giladreti for contributing!

PR #103, opened 2021-09-12
Add lighter Dockerfile to build those html and pdf in docker

25077667:
The original docker image will generate lots of middles-files that are owned by the root.
Alternatively, we can copy those "HTML"s and "PDF" from the image which is built from this Dockerfile directly.
By these commands:
```
id=$(docker create <IMAGE_NAME>)
docker cp $id:/workdir .   
docker rm -v $id
```

This image is much lighter and has zero side-effect.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,15 @@
+FROM twtug/lkmpg as builder
+LABEL MAINTAINER="25077667"
+
+COPY . /workdir
+
+RUN make all && \
+	make html && \
+	rm -f *.dvi *.aux *.log *.ps *.out lkmpg.bbl lkmpg.blg lkmpg.lof lkmpg.toc
+
+FROM alpine
+WORKDIR /workdir
+
+COPY --from=builder /workdir/lkmpg.pdf /workdir
+COPY --from=builder /workdir/html /workdir
+
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,4 +12,3 @@ WORKDIR /workdir
 
 COPY --from=builder /workdir/lkmpg.pdf /workdir
 COPY --from=builder /workdir/html /workdir
-


Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+RUN make all && \
+	make html && \
+	rm -f *.dvi *.aux *.log *.ps *.out lkmpg.bbl lkmpg.blg lkmpg.lof lkmpg.toc
+
+FROM alpine
+WORKDIR /workdir
+
+COPY --from=builder /workdir/lkmpg.pdf /workdir
+COPY --from=builder /workdir/html /workdir
+
```
jserv, line 15: Remove the trailing blank line.

jserv:
I would defer to @Hsins .

Hsins:
Hi @25077667,

Although using [multiple stage builds](https://docs.docker.com/develop/develop-images/multistage-build/) is an approach to keep docker images small, it's unnecessary and inappropriate in our case. It help us splitting Dockerfile into multiple sections (stages) and we can involve more than one image in builds.

The original docker image offers a consistent environment for building document and HTML files. People use the [multi stage builds image](https://github.com/25077667/lkmpg/blob/master/Dockerfile) would only get a container with result files instead of an environment for compiling. Remember that the scenario to reduce docker image size by using [multiple stage builds](https://docs.docker.com/develop/develop-images/multistage-build/) is that **the result files are essential for running on production environment and the libraries are not in need after compiling**.

25077667:
Well, It is designed for "RELEASE".

We dockerize or containerize projects would also expect to reduce the side-effects to the origin environment. And those documents and HTML files are also "productions", right?

I have no idea why it make lots of side-effect and I need to `rm` them with the `sudo` permission.
We don't need those middles files at the final stage at all (It was only needed in compile time).


> Hi @25077667,
> 
> Although using [multiple stage builds](https://docs.docker.com/develop/develop-images/multistage-build/) is an approach to keep docker images small, it's unnecessary and inappropriate in our case. It help us splitting Dockerfile into multiple sections (stages) and we can involve more than one image in builds.
> 
> The original docker image offers a consistent environment for building document and HTML files. People use the [multi stage builds image](https://github.com/25077667/lkmpg/blob/master/Dockerfile) would only get a container with result files instead of an environment for compiling. Remember that the scenario to reduce docker image size by using [multiple stage builds](https://docs.docker.com/develop/develop-images/multistage-build/) is that **the result files are essential for running on production environment and the libraries are not in need after compiling**.

Dockerfile
25077667: @jserv Line 12 and/or 15?

jserv: "trailing" means the end of this file.

Hsins:
> Well, It is designed for "RELEASE".
> 
> We dockerize or containerize projects would also expect to reduce the side-effects to the origin environment. And those documents and HTML files are also "productions", right?
> 
> I have no idea why it make lots of side-effect and I need to `rm` them with the `sudo` permission.
> We don't need those middles files at the final stage at all (It was only needed in compile time).

Currently, the [Generate Document Workflow](https://github.com/sysprog21/lkmpg/blob/master/.github/workflows/generate_doc.yaml) deals with the staff about release. It seems that the middle files would be deleted during that lifecycle and make no side-effects to origin environment. In other words:

- The middle files should be keep and may be useful for who would compile and debug with the container.
- For GitHub Actions, it's no need to care about the middle files. Those files would not be packaged inside the release.

25077667:
> > Well, It is designed for "RELEASE".
> > We dockerize or containerize projects would also expect to reduce the side-effects to the origin environment. And those documents and HTML files are also "productions", right?
> > I have no idea why it make lots of side-effect and I need to `rm` them with the `sudo` permission.
> > We don't need those middles files at the final stage at all (It was only needed in compile time).
> 
> Currently, the [Generate Document Workflow](https://github.com/sysprog21/lkmpg/blob/master/.github/workflows/generate_doc.yaml) deals with the stuff about release. It seems that the middle files would be deleted during that lifecycle and make no side-effects to origin environment (GitHub Action will pull that image and create container each time). In other words:
> 
> * The middle files should be keep and may be useful for who would compile and debug with the container.
> * For GitHub Actions, it's no need to care about the middle files. Those files would not be packed into release.

So, should we modify the README.md to say that docker image is just for CI?

Hsins:
> So, should we modify the README.md to say that docker image is just for CI?

As far as I am concerned, the adding dockerfile would not be helpful for current GitHub Actions workflow or user who wants to compile document in the container. It's weird to take those HTML and pdf files as the so-called "production **environment**" because we just pack those files for release instead of running them for serving.

Could you explain the benefits of it?

25077667:
> > So, should we modify the README.md to say that docker image is just for CI?
> 
> As far as I am concerned, the adding dockerfile would not be helpful for current GitHub Actions workflow or user who wants to compile document in the container. It's weird to take those HTML and pdf files as the so-called "production **environment**" because we just pack those files for release instead of running them for serving.
> 
> Could you explain the benefits of it?

The benefit is users can build those files by themself, as the released files, and they get no side effects.

Hsins:
> The benefit is users can build those files by themself, as the released files, and they get no side effects.

This reason failed to persuade me.

- [Those middle files are a natural byproduct when compiling documents and contain essential information needed for debugging or tracking labels](https://latex.org/forum/viewtopic.php?t=31921).
- People who use the original image for compiling and debugging need those files. And they can remove those files on the fly when it is in need.
- For the release workflow in GitHub Actions, it's no need to care about the middle files because we would not pack them into release.
- The possible scenario that people use the images using multiple stage builds for final document is rare.
    - People who want to contribute will use the original image for compiling and debugging.
    - People who want to download the compiled document can browse [the release page](https://github.com/sysprog21/lkmpg/releases) for downloading.

25077667:
> > The benefit is users can build those files by themself, as the released files, and they get no side effects.
> 
> This reason failed to persuade me.
> 
> * [Those middle files are a natural byproduct when compiling documents and contain essential information needed for debugging or tracking labels](https://latex.org/forum/viewtopic.php?t=31921).
> * People who use the original image for compiling and debugging need those files. And they can remove those files on the fly when it is in need.
> * For the release workflow in GitHub Actions, it's no need to care about the middle files because we would not pack them into release.
> * The possible scenario that people use the images using multiple stage builds for final document is rare.
>   
>   * People who want to contribute will use the original image (or their own environment) for compiling and debugging.
>   * People who want to download the compiled document can browse [the release page](https://github.com/sysprog21/lkmpg/releases) for downloading directly.

Okay, thanks.

PR #106, opened 2021-09-15
CI: Integrate the Sparse into the CI pipeline

linD026:
Sparse[1] is a semantic parser for C language, it can find out the
potential problem of the example code.

[1] https://www.kernel.org/doc/html/latest/dev-tools/sparse.html

### DIFF
--- a/.ci/static-analysis.sh
+++ b/.ci/static-analysis.sh
@@ -1,32 +1,65 @@
 #!/usr/bin/env bash
 
-SOURCES=$(find $(git rev-parse --show-toplevel) | egrep "\.(cpp|cc|c|h)\$")
+function do_cppcheck()
+{
+    local SOURCES=$(find $(git rev-parse --show-toplevel) | egrep "\.(cpp|cc|c|h)\$")
 
-CPPCHECK=$(which cppcheck)
-if [ $? -ne 0 ]; then
-    echo "[!] cppcheck not installed. Failed to run static analysis the source code." >&2
-    exit 1
-fi
+    local CPPCHECK=$(which cppcheck)
+    if [ $? -ne 0 ]; then
+        echo "[!] cppcheck not installed. Failed to run static analysis the source code." >&2
+        exit 1
+    fi
 
-## Suppression list ##
-# This list will explain the detail of suppressed warnings.
-# The prototype of the item should be like:
-# "- [{file}] {spec}: {reason}"
-#
-# - [hello-1.c] unusedFunction: False positive of init_module and cleanup_module.
-# - [*.c] missingIncludeSystem: Focus on the example code, not the kernel headers.
+    ## Suppression list ##
+    # This list will explain the detail of suppressed warnings.
+    # The prototype of the item should be like:
+    # "- [{file}] {spec}: {reason}"
+    #
+    # - [hello-1.c] unusedFunction: False positive of init_module and cleanup_module.
+    # - [*.c] missingIncludeSystem: Focus on the example code, not the kernel headers.
 
-OPTS="  --enable=warning,style,performance,information
-        --suppress=unusedFunction:hello-1.c
-        --suppress=missingIncludeSystem
-        --std=c89 "
+    local OPTS="
+            --enable=warning,style,performance,information
+            --suppress=unusedFunction:hello-1.c
+            --suppress=missingIncludeSystem
+            --std=c89 "
 
-$CPPCHECK $OPTS --xml ${SOURCES} 2> cppcheck.xml
-ERROR_COUNT=$(cat cppcheck.xml | egrep -c "</error>" )
+    $CPPCHECK $OPTS --xml ${SOURCES} 2> cppcheck.xml
+    local ERROR_COUNT=$(cat cppcheck.xml | egrep -c "</error>" )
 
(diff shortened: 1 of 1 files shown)


.ci/build-n-run.sh
```diff
@@ -26,5 +21,4 @@ function run_examples()
     done
 }
 
-build_example
```
jserv, line 29: Don't do this! We still have to ensure the default build rules.

.ci/sparse.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
+
+git clone git://git.kernel.org/pub/scm/devel/sparse/sparse.git
```
jserv, line 3: Don't use git. It takes time. Use tarball instead: http://www.kernel.org/pub/software/devel/sparse/dist/sparse-latest.tar.gz

jserv: See https://kernelnewbies.org/Sparse/

.ci/sparse.sh
```diff
@@ -0,0 +1,26 @@
...
+cd sparse
```
jserv, line 4: Use `pushd` and `popd`.

.ci/sparse.sh
```diff
@@ -0,0 +1,26 @@
...
+make
+sudo cp sparse /usr/bin/sparse
+cd ..
+
+SPARSE=$(which sparse)
```
jserv, line 9: Remove the check. Instead, you should validate the compilation of `sparse`.

.ci/sparse.sh
```diff
@@ -0,0 +1,26 @@
...
+WARNING_COUNT=$(cat sparse.log | egrep -c " warning:" )
+ERROR_COUNT=$(cat sparse.log | egrep -c " error:" )
+COUNT=`expr $WARNING_COUNT + $ERROR_COUNT`
+if [ $COUNT -gt 0 ]; then
+    echo "sparse failed: warning count is $WARNING_COUNT"
+    echo "sparse failed: error count is $ERROR_COUNT"
+    cat sparse.log
+    exit 1
+fi
+exit 0
```
jserv, line 26: Run `make clean` before the next step.

.ci/sparse.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
+
+wget http://www.kernel.org/pub/software/devel/sparse/dist/sparse-latest.tar.gz
```
jserv, line 3: Use `wget -q` to suppress the unintended messages.

.ci/sparse.sh
```diff
@@ -0,0 +1,26 @@
...
+if [ $? -ne 0 ]; then
+    echo "wget http://www.kernel.org/pub/software/devel/sparse/dist/sparse-latest.tar.gz failed"
```
jserv, line 5: Rewrite as
> Failed to download sparse.

.ci/sparse.sh
```diff
@@ -0,0 +1,26 @@
...
+    exit 1
+fi
+tar -xzf sparse-latest.tar.gz
+pushd sparse-*/
+make || exit 1
```
jserv, line 10: Rewrite as
```shell
make PREFIX=/usr all install || exit 1
```

.ci/sparse.sh
```diff
@@ -0,0 +1,26 @@
...
+sudo cp sparse /usr/bin/sparse
```
jserv, line 11: This command is not necessary.

.ci/sparse.sh
```diff
@@ -0,0 +1,26 @@
...
+sudo make PREFIX=/usr install || exit 1
+popd
+
+make -C examples C=2 2> sparse.log
+
+WARNING_COUNT=$(cat sparse.log | egrep -c " warning:" )
+ERROR_COUNT=$(cat sparse.log | egrep -c " error:" )
+COUNT=`expr $WARNING_COUNT + $ERROR_COUNT`
+if [ $COUNT -gt 0 ]; then
+    echo "sparse failed: warning count is $WARNING_COUNT"
```
jserv, line 20: The message can be as
> 1 warning(s), 2 error(s)

.ci/sparse.sh
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
+
+wget -q http://www.kernel.org/pub/software/devel/sparse/dist/sparse-latest.tar.gz
+if [ $? -ne 0 ]; then
+    echo "Failed to download sparse."
+    exit 1
+fi
+tar -xzf sparse-latest.tar.gz
+pushd sparse-*/
+make || exit 1
```
jserv, line 10: Change `make ` to `make sparse` to speed up build process.

.ci/sparse.sh
```diff
@@ -0,0 +1,25 @@
...
+sudo make PREFIX=/usr install || exit 1
```
jserv, line 11: Change the command to
```shell
sudo make INST_PROGRAMS=sparse PREFIX=/usr install
```

jserv (changes requested):
Merge the script ".ci/sparse.sh" into ".ci/static-analysis" since they have the same approach to validate the source code before executing.

PR #107, opened 2021-09-15
Fixed parameter name for hello-5.

sprowell:
Parameters are case-sensitive, so `myintArray` is different from `myintarray`, and people who are running the example with copy/paste will find it doesn't work as described in the text.  It is all lower-case in the code, so adjusted the example runs to match.


jserv:
Thank @sprowell for the insightful bug fix!

Issue #93, opened 2021-09-05
Several race conditions in chardev.c?


afiskon:
I believe there are several race conditions in chardev.c. The `msg` buffer is reused for several instances of the open file. If multiple processes will slowly read the same file this will cause memory corruption. Also, it seems like `open_device_cnt` should be atomic.

For simplicity, the best solution is probably to use a constant message. Also, I'm not an expert in Linux kernel modules. If there are in fact no bugs in the code, I think the text should explain in a little detail why this is the case.

afiskon:
@jserv  respectfully, the problem with:

```
static char msg[BUF_LEN];
```

... in chardev.c still exists. As I understand, `device_*` can be called concurrently which will cause concurrent access to `msg` and memory corruption as result. Unless there is something I don't know about how Linux kernel modules work. In the latest case I believe it should be clarified in the text and in the comments to the code.

jserv:
Reopen as per @afiskon requested. Under investigation.

YLowy:
I am testing the chardev.c on my machine and I think there is still some bug need to be fixed.

In my testing:
`[0] read 0 bytes.  
[1] read 0 bytes.  
[3] read 40 bytes. I already told you 1 times Hello world!
 
[2] read 40 bytes. I already told you 1 times Hello world!
 
[4] read 0 bytes.  
[3] read 40 bytes. I already told you 1 times Hello world!
 
[0] read 0 bytes.  
[1] read 40 bytes. I already told you 1 times Hello world!
 
[2] read 0 bytes.  
[4] read 40 bytes. I already told you 1 times Hello world!
`
Sometime, the read function from user space will fail (can't get the message) because the "Offset" in the open file table is not correct.
They all use the same open file table entry so there will also have race condition if we update the offset before device_read() return the bytes_read.

jserv:
> I am testing the chardev.c on my machine and I think there is still some bug need to be fixed.

@YLowy, you should discussion in the thread of pull request #108 if you were referring to the `offset` specific changes.

YLowy:
Due to the offset which is hard to be managed, I think it is a good way to create a list for each reading tasks. The element of the list include the task's PID and it's offset. So that the race condition in the open file table will not occur. Nevertheless, we still need to consider the list behavior when task close or lseek the device.

jserv:
Close via #108

PR #108, opened 2021-09-18
Add the description about some file ops and fix several problems

linD026:
Even though the concurrent access to the module has been fixed at commit 148fb01,
the function of device operations can be called concurrently.

In chardev2.c, the message buffer can be simultaneously accessed by the read and
write operations. It can solve by allowing only one of the operations to work at
the time.
In chardev.c, the msg_ptr is the global variable. Therefore, when someone is
concurrently using the read operation will lead to a data race.
The pointer can change into the local variable of the read operation. So, each
read operation will not use the same pointer at the same time.

---
Edited:

Since Linux v3.14, the read, write and seek operations of struct files are
guaranteed for thread safety [1][2]. This patch adds an explanation.

Following are the trivial problems:
- chardev.c
  - Move the "msg_ptr" pointer into the read function to remove unnecessary usage.
  - List the clear states of "already_open" by using mnemonic enumeration.

- chardev2.c
  - The "buffer" in the write function is user space data. It cannot use in the kernel space.
  - Reduce the redundant type transformation.
  - List the states of "already_open". Same as chardev.c.

[1] https://lore.kernel.org/lkml/20140303210359.26624.qmail@science.horizon.com/T/#u
[2] torvalds/linux@9c225f2

- sysprog21/lkmpg: https://github.com/sysprog21/lkmpg/issues/93

### DIFF
--- a/examples/chardev.c
+++ b/examples/chardev.c
@@ -27,10 +27,16 @@ static ssize_t device_write(struct file *, const char __user *, size_t,
 /* Global variables are declared as static, so are global within the file. */
 
 static int major; /* major number assigned to our device driver */
+
+enum {
+    CDEV_NOT_USED = 0,
+    CDEV_EXCLUSIVE_OPEN = 1,
+};
+
 /* Is device open? Used to prevent multiple access to device */
-static atomic_t already_open = ATOMIC_INIT(0);
+static atomic_t already_open = ATOMIC_INIT(CDEV_NOT_USED);
+
 static char msg[BUF_LEN]; /* The msg the device will give when asked */
-static char *msg_ptr;
 
 static struct class *cls;
 
@@ -78,11 +84,10 @@ static int device_open(struct inode *inode, struct file *file)
 {
     static int counter = 0;
 
-    if (atomic_cmpxchg(&already_open, 0, 1))
+    if (atomic_cmpxchg(&already_open, CDEV_NOT_USED, CDEV_EXCLUSIVE_OPEN))
         return -EBUSY;
 
     sprintf(msg, "I already told you %d times Hello world!\n", counter++);
-    msg_ptr = msg;
     try_module_get(THIS_MODULE);
 
     return SUCCESS;
@@ -91,7 +96,8 @@ static int device_open(struct inode *inode, struct file *file)
 /* Called when a process closes the device file. */
 static int device_release(struct inode *inode, struct file *file)
 {
-    atomic_set(&already_open, 0); /* We're now ready for our next caller */
+    /* We're now ready for our next caller */
+    atomic_set(&already_open, CDEV_NOT_USED);
 
     /* Decrement the usage count, or else once you opened the file, you will
      * never get get rid of the module.
@@ -111,10 +117,14 @@ static ssize_t device_read(struct file *filp, /* see include/linux/fs.h   */
 {
     /* Number of bytes actually written to the buffer */
     int bytes_read = 0;
+    const char *msg_ptr = msg;
 
-    /* If we are at the end of message, return 0 signifying end of file. */
-    if (*msg_ptr == 0)
-        return 0;
+    if (!*(msg_ptr + *offset)) { /* we are at the end of message */
+        *offset = 0; /* reset the offset */
+        return 0; /* signify end of file */
+    }
+
+    msg_ptr += *offset;
 
     /* Actually put the data into the buffer */
     while (length && *msg_ptr) {
@@ -124,10 +134,11 @@ static ssize_t device_read(struct file *filp, /* see include/linux/fs.h   */
          * the user data segment.
          */
         put_user(*(msg_ptr++), buffer++);
-
         length--;
         bytes_read++;
     }
+
+    *offset += bytes_read;
 
     /* Most read functions return the number of bytes put into the buffer. */
     return bytes_read;
--- a/examples/chardev2.c
+++ b/examples/chardev2.c
@@ -17,18 +17,18 @@
 #define DEVICE_NAME "char_dev"
 #define BUF_LEN 80
 
+enum {
+    CDEV_NOT_USED = 0,
+    CDEV_EXCLUSIVE_OPEN = 1,
+};
+
 /* Is the device open right now? Used to prevent concurrent access into
  * the same device
  */
-static atomic_t already_open = ATOMIC_INIT(0);
+static atomic_t already_open = ATOMIC_INIT(CDEV_NOT_USED);
 
 /* The message the device will give when asked */
 static char message[BUF_LEN];
-
-/* How far did the process reading the message get? Useful if the message
- * is larger than the size of the buffer we get to fill in device_read.
- */
-static char *message_ptr;
 
 static struct class *cls;
 
@@ -38,11 +38,9 @@ static int device_open(struct inode *inode, struct file *file)
     pr_info("device_open(%p)\n", file);
 
     /* We don't want to talk to two processes at the same time. */
-    if (atomic_cmpxchg(&already_open, 0, 1))
+    if (atomic_cmpxchg(&already_open, CDEV_NOT_USED, CDEV_EXCLUSIVE_OPEN))
         return -EBUSY;
 
-    /* Initialize the message */
-    message_ptr = message;
     try_module_get(THIS_MODULE);
     return SUCCESS;
 }
@@ -52,7 +50,7 @@ static int device_release(struct inode *inode, struct file *file)
     pr_info("device_release(%p,%p)\n", inode, file);
 
     /* We're now ready for our next caller */
-    atomic_set(&already_open, 0);
+    atomic_set(&already_open, CDEV_NOT_USED);
 
     module_put(THIS_MODULE);
     return SUCCESS;
@@ -68,12 +66,17 @@ static ssize_t device_read(struct file *file, /* see include/linux/fs.h   */
 {
     /* Number of bytes actually written to the buffer */
     int bytes_read = 0;
+    /* How far did the process reading the message get? Useful if the message
+     * is larger than the size of the buffer we get to fill in device_read.
+     */
+    const char *message_ptr = message;
 
-    pr_info("device_read(%p,%p,%ld)\n", file, buffer, length);
+    if (!*(message_ptr + *offset)) { /* we are at the end of message */
+        *offset = 0; /* reset the offset */
+        return 0; /* signify end of file */
+    }
 
-    /* If at the end of message, return 0 (which signifies end of file). */
-    if (*message_ptr == 0)
-        return 0;
+    message_ptr += *offset;
 
     /* Actually put the data into the buffer */
     while (length && *message_ptr) {
@@ -89,6 +92,8 @@ static ssize_t device_read(struct file *file, /* see include/linux/fs.h   */
 
     pr_info("Read %d bytes, %ld left\n", bytes_read, length);
 
+    *offset += bytes_read;
+
     /* Read functions are supposed to return the number of bytes actually
      * inserted into the buffer.
      */
@@ -101,12 +106,10 @@ static ssize_t device_write(struct file *file, const char __user *buffer,
 {
     int i;
 
-    pr_info("device_write(%p,%s,%ld)", file, buffer, length);
+    pr_info("device_write(%p,%p,%ld)", file, buffer, length);
 
     for (i = 0; i < length && i < BUF_LEN; i++)
         get_user(message[i], buffer + i);
-
-    message_ptr = message;
 
     /* Again, return the number of input characters used. */
     return i;
@@ -126,43 +129,44 @@ device_ioctl(struct file *file, /* ditto */
              unsigned long ioctl_param)
 {
     int i;
-    char *temp;
-    char ch;
 
     /* Switch according to the ioctl called */
     switch (ioctl_num) {
-    case IOCTL_SET_MSG:
+    case IOCTL_SET_MSG: {
         /* Receive a pointer to a message (in user space) and set that to
-         * be the device's message.  Get the parameter given to ioctl by
+         * be the device's message. Get the parameter given to ioctl by
          * the process.
          */
-        temp = (char *)ioctl_param;
+        char __user *tmp = (char __user *)ioctl_param;
+        char ch;
 
         /* Find the length of the message */
-        get_user(ch, (char __user *)temp);
-        for (i = 0; ch && i < BUF_LEN; i++, temp++)
-            get_user(ch, (char __user *)temp);
+        get_user(ch, tmp);
+        for (i = 0; ch && i < BUF_LEN; i++, tmp++)
+            get_user(ch, tmp);
 
         device_write(file, (char __user *)ioctl_param, i, NULL);
         break;
+    }
+    case IOCTL_GET_MSG: {
+        loff_t offset = 0;
 
-    case IOCTL_GET_MSG:
         /* Give the current message to the calling process - the parameter
          * we got is a pointer, fill it.
          */
-        i = device_read(file, (char __user *)ioctl_param, 99, NULL);
+        i = device_read(file, (char __user *)ioctl_param, 99, &offset);
 
         /* Put a zero at the end of the buffer, so it will be properly
          * terminated.
          */
         put_user('\0', (char __user *)ioctl_param + i);
         break;
-
+    }
     case IOCTL_GET_NTH_BYTE:
         /* This ioctl is both input (ioctl_param) and output (the return
          * value of this function).
          */
-        return message[ioctl_param];
+        return (long)message[ioctl_param];
         break;
     }
 
--- a/lkmpg.tex
+++ b/lkmpg.tex
@@ -835,6 +835,9 @@ The meaning is clear, and you should be aware that any member of the structure w
 
 An instance of \cpp|struct file_operations| containing pointers to functions that are used to implement \cpp|read|, \cpp|write|, \cpp|open|, \ldots{} system calls is commonly named \cpp|fops|.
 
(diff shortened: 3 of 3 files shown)


examples/chardev.c
```diff
@@ -111,11 +109,14 @@ static ssize_t device_read(struct file *filp, /* see include/linux/fs.h   */
 {
     /* Number of bytes actually written to the buffer */
     int bytes_read = 0;
+    char *msg_ptr = msg;
 
     /* If we are at the end of message, return 0 signifying end of file. */
-    if (*msg_ptr == 0)
+    if (strlen(msg) == *offset)
```
jserv, line 115: Is it possible to eliminate the use of `strlen` here?

linD026:
According to [this comment](https://github.com/sysprog21/lkmpg/issues/93#issuecomment-922769820):

> Sometime, the read function from user space will fail (can't get the message) because the "Offset" in the open file table is not correct.
They all use the same open file table entry so there will also have race condition if we update the offset before device_read() return the bytes_read.

With multi-read, the "offset" variable will have a race condition.
Do we need to deal with it?
If so, the read function may become like:
```cpp
char *msg_ptr = msg;

while (length && *msg_ptr) {
    put_user(*(msg_ptr++), buffer++);
    length--;
}

return 0;
```

examples/chardev.c
```diff
@@ -111,10 +109,15 @@ static ssize_t device_read(struct file *filp, /* see include/linux/fs.h   */
 {
     /* Number of bytes actually written to the buffer */
     int bytes_read = 0;
+    char *msg_ptr = msg;
```
jserv, line 112: Can you add `const` qualifier?

examples/chardev2.c
```diff
@@ -68,12 +61,23 @@ static ssize_t device_read(struct file *file, /* see include/linux/fs.h   */
...
+        *offset = 0;
+        return 0;
+    }
 
     pr_info("device_read(%p,%p,%ld)\n", file, buffer, length);
 
-    /* If at the end of message, return 0 (which signifies end of file). */
-    if (*message_ptr == 0)
-        return 0;
+    if (atomic_cmpxchg(&already_open, 1, 2) == 2)
```
jserv, line 79: `2` looks a bit tricky to understand.

examples/chardev.c
```diff
@@ -111,10 +109,15 @@ static ssize_t device_read(struct file *filp, /* see include/linux/fs.h   */
 {
     /* Number of bytes actually written to the buffer */
     int bytes_read = 0;
+    const char *msg_ptr = msg;
 
-    /* If we are at the end of message, return 0 signifying end of file. */
-    if (*msg_ptr == 0)
+    /* If we are at the end of message, reset the offset and
```
jserv, line 114: Re-format the comment to wrap to 80-characters.

examples/chardev2.c
jserv: Also, we might have to update TeX document for the illustration of this part.

examples/chardev2.c
```diff
@@ -126,7 +143,8 @@ device_ioctl(struct file *file, /* ditto */
              unsigned long ioctl_param)
 {
     int i;
-    char *temp;
+    loff_t offset = 0;
+    char __user *temp;
```
jserv, line 147: Let's use `tmp` for short.

examples/chardev.c
```diff
@@ -111,10 +109,15 @@ static ssize_t device_read(struct file *filp, /* see include/linux/fs.h   */
 {
     /* Number of bytes actually written to the buffer */
     int bytes_read = 0;
+    const char *msg_ptr = msg;
 
-    /* If we are at the end of message, return 0 signifying end of file. */
-    if (*msg_ptr == 0)
+    /* If we are at the end of message, reset the offset and return 0 (which
```
jserv, line 114: Rewrite as following:
```c
    if (!*(msg_ptr + *offset)) { /* we are at the end of message */
        *offset = 0; /* reset the offset */
        return 0; /* signify end of file */
    }
```

For writing purpose, we would stick to shorter code snip.

examples/chardev2.c
```diff
@@ -136,12 +154,12 @@ device_ioctl(struct file *file, /* ditto */
          * be the device's message.  Get the parameter given to ioctl by
          * the process.
          */
-        temp = (char *)ioctl_param;
+        tmp = (char __user *)ioctl_param;
```
jserv, line 157: Declare `tmp` here. That is,
```c
    case IOCTL_SET_MSG: {
        /* Receive a pointer to a message (in user space) and set that to
         * be the device's message.  Get the parameter given to ioctl by
         * the process.
         */
        char __user *tmp = (char __user *)ioctl_param;
        ...
    }
```

linD026: It may lead Sparse report the error:
```shell
/lkmpg/examples/chardev2.c: In function ‘device_ioctl’:
/lkmpg/examples/chardev2.c:153:9: error: a label can only be part of a statement and a declaration is not a statement
  153 |         char __user *tmp = (char __user *)ioctl_param;
      |         ^~~~

```

linD026: Oh, I found the example here: [linux/arch/alpha/kernel/smc37c669.c](https://github.com/torvalds/linux/blob/master/arch/alpha/kernel/smc37c669.c#L1233)
It seems fine. I will change it.

linD026: It may lead the Cppcheck report error.
```shell
        <error id="variableScope" severity="style" msg="The scope of the variable &apos;ch&apos; can be reduced." verbose="The scope of the variable &apos;ch&apos; can be reduced. Warning: Be careful when fixing this message, especially when there are inner loops. Here is an example where cppcheck will write that the scope for &apos;i&apos; can be reduced:\012void f(int x)\012{\012    int i = 0;\012    if (x) {\012        // it&apos;s safe to move &apos;int i = 0;&apos; here\012        for (int n = 0; n &lt; 10; ++n) {\012            // it is possible but not safe to move &apos;int i = 0;&apos; here\012            do_something(&amp;i);\012        }\012    }\012}\012When you see this message it is always safe to reduce the variable scope 1 level." cwe="398">
            <location file="/home/slda/lkmpg/examples/chardev2.c" line="143" column="10"/>
            <symbol>ch</symbol>
        </error>
```

jserv: We can safely suppress the warning "The scope of the variable ??? can be reduced" since it does not harm.

linD026: > It may lead the Cppcheck report error.
> 
> ```shell
>         <error id="variableScope" severity="style" msg="The scope of the variable &apos;ch&apos; can be reduced." verbose="The scope of the variable &apos;ch&apos; can be reduced. Warning: Be careful when fixing this message, especially when there are inner loops. Here is an example where cppcheck will write that the scope for &apos;i&apos; can be reduced:\012void f(int x)\012{\012    int i = 0;\012    if (x) {\012        // it&apos;s safe to move &apos;int i = 0;&apos; here\012        for (int n = 0; n &lt; 10; ++n) {\012            // it is possible but not safe to move &apos;int i = 0;&apos; here\012            do_something(&amp;i);\012        }\012    }\012}\012When you see this message it is always safe to reduce the variable scope 1 level." cwe="398">
>             <location file="/lkmpg/examples/chardev2.c" line="143" column="10"/>
>             <symbol>ch</symbol>
>         </error>
> ```

Fixed by moving `char ch` into the scope.
```shell
-    char ch;
 
     /* Switch according to the ioctl called */
     switch (ioctl_num) {
@@ -151,6 +149,7 @@ device_ioctl(struct file *file, /* ditto */
          * the process.
          */
         char __user *tmp = (char __user *)ioctl_param;
+        char ch;
```

examples/chardev2.c
jserv: Again, we never rely on the magic numbers for this case.

jserv:
I would like to ask @afiskon for reviewing.

examples/chardev.c
```diff
@@ -27,10 +27,16 @@ static ssize_t device_write(struct file *, const char __user *, size_t,
 /* Global variables are declared as static, so are global within the file. */
 
 static int major; /* major number assigned to our device driver */
+
 /* Is device open? Used to prevent multiple access to device */
 static atomic_t already_open = ATOMIC_INIT(0);
+
+enum {
```
jserv, line 34: Don't do that. Each constant in this `enum` should have a common prefix. Let's say, `CDEV_`. Also, shorten the names of constants.

lkmpg.tex
```diff
@@ -927,8 +927,8 @@ \subsection{chardev.c}
 In the multiple-threaded environment, without any protection, concurrent access to the same memory may lead to the race condition, and will not preserve the performance.
 In the kernel module, this problem may happen due to multiple instances accessing the shared resources.
 Therefore, a solution is to enforce the exclusive access.
-We use atomic Compare-And-Swap (CAS), the single atomic operation, to determine whether the file is currently open by someone.
-CAS compares the contents of a memory loaction with the expected value and, only if they are the same, modifies the contents of that memory location to the desired value.
+We use atomic Compare-And-Swap (CAS), the single atomic operation, to determine whether the file is currently open by someone and the shared resource is access simultaneously.
```
jserv, line 930: Remove "the single atomic operation" which looks redundant.

examples/chardev.c
```diff
@@ -27,10 +27,16 @@ static ssize_t device_write(struct file *, const char __user *, size_t,
 /* Global variables are declared as static, so are global within the file. */
 
 static int major; /* major number assigned to our device driver */
+
+enum {
+    CDEV_UNUSED = 0,
```
jserv, line 32: Replace `CDEV_UNUSED` to `CDEC_NOT_USED` for being symmetric.

lkmpg.tex
```diff
@@ -927,8 +927,8 @@ \subsection{chardev.c}
 In the multiple-threaded environment, without any protection, concurrent access to the same memory may lead to the race condition, and will not preserve the performance.
 In the kernel module, this problem may happen due to multiple instances accessing the shared resources.
 Therefore, a solution is to enforce the exclusive access.
-We use atomic Compare-And-Swap (CAS), the single atomic operation, to determine whether the file is currently open by someone.
-CAS compares the contents of a memory loaction with the expected value and, only if they are the same, modifies the contents of that memory location to the desired value.
+We use atomic Compare-And-Swap (CAS) to determine whether the file is currently open by someone and the shared resource is access simultaneously.
```
jserv, line 930: Mention `CDEV_EXCLUSIVE_OPEN` and/or `CDEV_EXCLUSIVE_ACCESS` in TeX document.

linD026:
> From this [mail list](https://lore.kernel.org/lkml/53022DB1.4070805@gmail.com/), when the reference count of `struct file` (`files->count`) is larger than 1, it will atomically do the {read, write, lseek} syscall[[1]](https://github.com/torvalds/linux/blob/78e709522d2c012cb0daad2e668506637bffb7c2/fs/file.c#L982)[[2]](https://github.com/torvalds/linux/blob/master/fs/read_write.c#L612).
> 
> * [file.c](https://elixir.bootlin.com/linux/latest/source/fs/file.c#L966)
> 
> ```c++
> unsigned long __fdget_pos(unsigned int fd)
> {
> 	unsigned long v = __fdget(fd);
> 	struct file *file = (struct file *)(v & ~3);
> 
> 	if (file && (file->f_mode & FMODE_ATOMIC_POS)) {
> 		if (file_count(file) > 1) {
> 			v |= FDPUT_POS_UNLOCK;
> 			mutex_lock(&file->f_pos_lock);
> 		}
> 	}
> 	return v;
> }
> ```
> 
> And, the `f_pos` variable will not have the race condition, which is the stored offset value in the `struct file`.
> 
> * [fs.h](https://elixir.bootlin.com/linux/latest/source/include/linux/fs.h#L920)
> 
> ```c++
> struct file {
> 	loff_t			f_pos;
> }
> ```
> 
> * [read_write.c](https://github.com/torvalds/linux/blob/master/fs/read_write.c#L612)
> 
> ```c++
> ssize_t ksys_read(unsigned int fd, char __user *buf, size_t count)
> {
> 	struct fd f = fdget_pos(fd);
> 	ssize_t ret = -EBADF;
> 
> 	if (f.file) {
> 		loff_t pos, *ppos = file_ppos(f.file);
> 		if (ppos) {
> 			pos = *ppos;
> 			ppos = &pos;
> 		}
> 		ret = vfs_read(f.file, buf, count, ppos);
> 		if (ret >= 0 && ppos)
> 			f.file->f_pos = pos;
> 		fdput_pos(f);
> 	}
> 	return ret;
> }
> ```
> 
> So, do we have to change the example code?

linD026:
I tested it.
With the following changes, It seems like the read and write functions will work flawlessly.
```cpp
static ssize_t device_read(struct file *filp, /* see include/linux/fs.h   */
                           char __user *buffer, /* buffer to fill with data */
                           size_t length, /* length of the buffer     */
                           loff_t *offset)
{
    int bytes_read = 0;
    const char *msg_ptr = msg;

    if (!*(msg_ptr + *offset)) { /* we are at the end of message */
        *offset = 0; /* reset the offset */
        return 0; /* signify end of file */
    }

    msg_ptr += *offset;

    while (length && *msg_ptr) {
        put_user(*(msg_ptr++), buffer++);

        length--;
        bytes_read++;
    }

    *offset += bytes_read;

    return bytes_read;
}
```
test code:
```cpp
#define _GNU_SOURCE
#include <stdio.h>
#include <fcntl.h>
#include <unistd.h>
#include <pthread.h>

int fd;

#define len 10

void worker(void *args)
{
    char buf[len];

    read(fd, buf, len);
    printf("unlock %d: \"%s\"\n", gettid(), buf);

    pthread_exit(NULL);
}

int main(int argc, char *argv[])
{
    pthread_t t[10];
    
    fd = open("/dev/chardev", O_RDWR);

    for (int i = 0;i < 10;i++)
        pthread_create(&t[i], NULL, worker, NULL);
    
    for (int i = 0;i < 10;i++)
        pthread_join(t[i], NULL);
    
    return 0;
}

```
output (it is sequential output):
```shell
unlock 15733: "I already "
unlock 15732: "told you 2"
unlock 15734: " times Hel"
unlock 15735: "lo world!
"
unlock 15731: ""
unlock 15736: "I already "
unlock 15730: "told you 2"
unlock 15729: " times Hel"
unlock 15728: "lo world!
"
unlock 15727: ""
```

lkmpg.tex
```diff
@@ -927,8 +930,8 @@ \subsection{chardev.c}
 In the multiple-threaded environment, without any protection, concurrent access to the same memory may lead to the race condition, and will not preserve the performance.
 In the kernel module, this problem may happen due to multiple instances accessing the shared resources.
 Therefore, a solution is to enforce the exclusive access.
-We use atomic Compare-And-Swap (CAS), the single atomic operation, to determine whether the file is currently open by someone.
-CAS compares the contents of a memory loaction with the expected value and, only if they are the same, modifies the contents of that memory location to the desired value.
+We use atomic Compare-And-Swap (CAS) to determine whether the file is currently open by someone.
```
fennecJ, line 933: should be `opened`.

examples/chardev2.c
```diff
@@ -68,12 +66,19 @@ static ssize_t device_read(struct file *file, /* see include/linux/fs.h   */
...
+        *offset = 0; /* reset the offset */
+        return 0; /* signify end of file */
+    }
 
-    /* If at the end of message, return 0 (which signifies end of file). */
-    if (*message_ptr == 0)
-        return 0;
+    message_ptr += *offset;
+
+    pr_info("device_read(%p,%p,%ld)\n", file, buffer, length);
```
jserv, line 81: This `printk` might not be useful.

lkmpg.tex
```diff
@@ -835,6 +835,9 @@ \subsection{The file\_operations Structure}
 
 An instance of \cpp|struct file_operations| containing pointers to functions that are used to implement \cpp|read|, \cpp|write|, \cpp|open|, \ldots{} system calls is commonly named \cpp|fops|.
 
+Since Linux v3.14, the read, write and seek operations are guaranteed for thread-safe by letting the file position update being mutual exclusion.
```
jserv, line 838: Avoid using `letting` since we need the subject and/or someone to perform the action.

lkmpg.tex
```diff
@@ -927,8 +930,8 @@ \subsection{chardev.c}
 In the multiple-threaded environment, without any protection, concurrent access to the same memory may lead to the race condition, and will not preserve the performance.
 In the kernel module, this problem may happen due to multiple instances accessing the shared resources.
 Therefore, a solution is to enforce the exclusive access.
-We use atomic Compare-And-Swap (CAS), the single atomic operation, to determine whether the file is currently open by someone.
-CAS compares the contents of a memory loaction with the expected value and, only if they are the same, modifies the contents of that memory location to the desired value.
+We use atomic Compare-And-Swap (CAS) to determine whether the file is currently opened by someone.
+CAS compares the contents of a memory location with the expected value and, only if they are the same, modifies the contents of that memory location to the desired value.
```
jserv, line 934: In TeX document, you should explain the use of atomic operations in the above LKM, at least for `CDEV_NOT_USED` and `CDEV_EXCLUSIVE_OPEN`.

lkmpg.tex
```diff
@@ -927,8 +930,8 @@ \subsection{chardev.c}
 In the multiple-threaded environment, without any protection, concurrent access to the same memory may lead to the race condition, and will not preserve the performance.
 In the kernel module, this problem may happen due to multiple instances accessing the shared resources.
 Therefore, a solution is to enforce the exclusive access.
-We use atomic Compare-And-Swap (CAS), the single atomic operation, to determine whether the file is currently open by someone.
-CAS compares the contents of a memory loaction with the expected value and, only if they are the same, modifies the contents of that memory location to the desired value.
+We use atomic Compare-And-Swap (CAS) to maintain the states, \cpp|CDEV_NOT_USED| and \cpp|CDEV_EXCLUSIVE_OPEN|, to determine the file is currently opened by someone or not.
```
jserv, line 933: Use "determine whether ... or not"

PR #110, opened 2021-09-21
Multi-read supported chardev.c without race condition of offset.

YLowy:


examples/chardev.c
```diff
@@ -13,6 +13,13 @@
 #include <linux/module.h>
 #include <linux/poll.h>
 
+/* Structure for every task's offset */
+typedef struct element{
+    int pid;
+    long offset;
+    struct element *next;
+}list_ele_t;
```
jserv, line 21: Run `clang-format -i chardev.c` in directory `examples`.

jserv (changes requested):
Follow Linux kernel coding style and avoid using camelCase such as `currentOffset`.

linD026:
This improvement with the multi-read is unnecessary because the read operation of the `struct file` already being atomic since Linux v3.14.
You can see more [here](https://github.com/sysprog21/lkmpg/pull/108#issuecomment-923965811).
Ps. Apologize for the previously misleading information that I talked about in the private messages.

jserv:
Close as per @linD026 suggested.

PR #113, opened 2021-09-23
Prepare to add the pull request status-check from GitHub

linD026:
We have to reject the pull request if the static analysis
of the CI pipeline has failed.

Related article:
https://www.wesleyhaakman.org/working-with-pull-requests-status-checks-arm-templates-and-github-actions/


jserv:
Can we check the job `validate` only? It would be much faster and make sense to contributions.

linD026:
Yes, it can.
Just add the `validate` to the status-check list.
But why the `validate` check can work without adding the `on: pull_request` statement?

linD026:
My test [here](https://github.com/linD026/GitHub-workflow-test/pull/1) cannot check without adding `on: pull_request`.
It will get the `out-of-date` or `Required statuses must pass before merging` warning.
> no-pr-validate Expected — Waiting for status to be reported Required

jserv:
> Just add the `validate` to the status-check list.
> But why the `validate` check can work without adding the `on: pull_request` statement?

Cc. @fennecJ
I tend to run the job `validate` only without any document building and deployment. That is, contributors who submit pull requests must be approved before the necessary coding style checks and static analysis.

linD026:
Can you just only select the `validate`?
Like the following:

linD026:
> > Just add the `validate` to the status-check list.
> > But why the `validate` check can work without adding the `on: pull_request` statement?
> 
> Cc. @fennecJ
> I tend to run the job `validate` only without any document building and deployment. That is, contributors who submit pull requests must be approved before the necessary coding style checks and static analysis.

I use the `if` statement to skip the `build`.
```yaml
if: ${{ github.ref == github.base_ref}}
```

jserv:
> I use the `if` statement to skip the `build`.
> ```yaml
> if: ${{ github.ref == github.base_ref}}
> ```

Splitting `validate` job from `.github/workflows/generate_doc.yaml` would improve modularity.

.github/workflows/status-check.yaml
```diff
@@ -0,0 +1,24 @@
+name: build-deploy-assets
```
fennecJ, line 1: The name should be changed.

fennecJ (changes requested):

fennecJ (approved):
Looks good to me.

PR #115, opened 2021-09-25
Update lkmpg.tex

demonsome:
I suggest that the explanation of "inode" should appear at the first place where reader will first see it.


jserv:
Please read the article [How to Write a Git Commit Message](https://chris.beams.io/posts/git-commit/) carefully and then change the subject/body of git commit message.

PR #116, opened 2021-09-25
Provide a tasklet-free example

jserv:
It is possible that in future tasklets may be replaced by threaded irqs.

### DIFF
--- a/lkmpg.tex
+++ b/lkmpg.tex
@@ -1570,6 +1570,10 @@ There are two main ways of running tasks: tasklets and work queues.
 Tasklets are a quick and easy way of scheduling a single function to be run.
 For example, when triggered from an interrupt, whereas work queues are more complicated but also better suited to running multiple things in a sequence.
 
+It is possible that in future tasklets may be replaced by \textit{threaded irqs}.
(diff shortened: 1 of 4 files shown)


lkmpg.tex
```diff
@@ -1669,6 +1673,11 @@ \subsection{Bottom Half}
 
 \samplec{examples/bottomhalf.c}
 
+It is expected that at some unspecified point in the future tasklets may eventually be deprecated, to be replaced by \textit{threaded irqs} instead.
```
linD026, line 1676: In 2020, the removal of the tasklet API also has been discussed.
See https://lwn.net/Articles/830964/.

linD026: Also, the patch of tasklets in 2020 has the same improvement as the `timer_setup`. In other words, the CET needs to be mention too.
See:
* https://github.com/torvalds/linux/commit/12cc923f1ccc1df467e046b02a72c2b3b321b6a2
* https://lwn.net/ml/kernel-hardening/20200716030847.1564131-1-keescook@chromium.org/

linD026 (changes requested):

0xff07:
I sent a pull request to the `threaded-irq` branch.

jserv:
Close in favor of #255

PR #119, opened 2021-09-29
Correct English gramma

marconi1964:
Correct English gramma

Original text : within "because at that time, the files you need will be are ready.",  "will be" is redundant.
Corrected text : "because at that time, the files you need are ready."


jserv:
Thank @marconi1964 for pointing this out.

PR #123, opened 2021-10-19
Generate contributor list as a modular inclusion

fennecJ:
This script do mainly 2 things:
1.
Generate file `Contributors` with git log in following format:
Contributor's name,<1-st e-mail>[,<2nd e-mail>][,<3rd e-mail>]...
The 2-nd email and so on are based on file `.mailmap` in the root
directory of the repo.
Note that it will also append contributors in File `Include`; and
will NOT append contributors in File `Exclude`.
If there are new contributors, script will sort `Contributor` after
append new contributors into each file; Otherwise it do nothing.

2.
Generate contrib.tex based on `Contributors` into lib/contrib.tex in
following format:
[name], (reasonable width) % [1-st e-mail]
Which is inspired by The Not So Short Introduction to Latex 2e[1]

We need to maintain `.mailmap`, `Exclude` and `Include` manually.
All Chinese name should be converted into English/Pinyin in `.mailmap`,
otherwise we may need extra pkg for latex to parse Chinese characters.

[1] - https://tobi.oetiker.ch/lshort/lshort.pdf

Close sysprog21#68

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -1,0 +1,8 @@
+JianXing Wu <fdgkhdkgh@gmail.com> 吳建興 <fdgkhdkgh@gmail.com>
+fennecJ     <hwahwa649@gmail.com> fennecJ   <58484289+fennecJ@users.noreply.github.com>
+fennecJ     <hwahwa649@gmail.com> fennecj   <hwahwa649@gmail.com>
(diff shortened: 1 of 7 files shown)


linD026:
Looks good!

PR #124, opened 2021-10-26
Fix grammar

tzuyichan:
in the hope ... will -> in the hope that ... would
get -> got
where are ... gone -> where have ... gone
For those not -> For those who are not

### DIFF
--- a/lkmpg.tex
+++ b/lkmpg.tex
@@ -55,7 +55,7 @@
 \label{sec:introduction}
 The Linux Kernel Module Programming Guide is a free book; you may reproduce and/or modify it under the terms of the \href{https://opensource.org/licenses/OSL-3.0}{Open Software License}, version 3.0.
 
-This book is distributed in the hope it will be useful, but without any warranty, without even the implied warranty of merchantability or fitness for a particular purpose.
+This book is distributed in the hope that it would be useful, but without any warranty, without even the implied warranty of merchantability or fitness for a particular purpose.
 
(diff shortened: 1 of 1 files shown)


tzuyichan:
Sure, I just did.
Unfortunately, Grammarly and Trinka had a hard time picking up on these subtler grammar patterns. I think mixing programming terms and variables in the text also affects their accuracy.

Lin2Jing4 (approved):

leewei05:
@tzuyichan @jserv 
Other than Grammarly and Trinka, we can also use Linggle or Netspeak to check if certain phrases are commonly used.
For example, we can search for the use of [`in the hope that`](https://netspeak.org/#q=in+the+hope+that).
Moreover, we can use `? or ...` to find one or many words, [`where have ? gone`](https://netspeak.org/#q=where+have+%3F+gone)

As for `For those not` and `For those who are not`, I personally think they are both correct. However, if you consider the coherence of these sentences, I would rewrite them as the following:

Original
> Those familiar with shell scripts will easily be able to spot them. For those not, the obj-$(CONFIG_FOO) entries you see everywhere expand into obj-y or obj-m, depending on whether the CONFIG_FOO variable has been set to y or m.

New
> Those (who are) familiar with shell scripts can easily know the answer. For those who are not, every obj-$(CONFIG_FOO) entries are expanded into obj-y or obj-m, depending on whether we set the CONFIG_FOO variable to y or m.

Because the last sentence was a question, I change the some of the wording.

jserv:
Thank @tzuyichan for contributing!

PR #125, opened 2021-10-31
CI: Evaluate Smatch for static analysis

### LINKED ISSUE #122: Evaluate smatch for static analysis
[smatch](https://github.com/error27/smatch) was built on sparse. We might introduce [smatch](https://github.com/error27/smatch) for static analysis.
See https://lwn.net/Articles/696624/

linD026:
Smatch[1][2] is a pluggable static analysis for C. It may help us find
out the potential problem of the example code.

Doing with smatch, if set the --file-output flag, it will generate the
{}.c.smatch report for each c file. This will make a little bit
complicated to collect all the report messages. So, here we stay at the
default setting, stdout for the smatch messages.

For more information, see:
- https://lwn.net/Articles/696624/
- https://elinux.org/images/d/d3/Bargmann.pdf

Also, fix the warning from Smatch:

Smatch failed: 1 warning(s), 1 error(s)
/home/runner/work/lkmpg/lkmpg/examples/procfs2.c:57 procfile_write() error: buffer overflow 'procfs_buffer' 1024 <= 1024
/home/runner/work/lkmpg/lkmpg/examples/kbleds.c:58 kbleds_init() warn: argument 5 to %lx specifier is cast from pointer

Futhermore, the effect of the write operation in procfs2.c is too
implied. So after writing, print the buffer every time.

Close #122

[1] https://github.com/error27/smatch
[2] https://repo.or.cz/w/smatch.git
* sysprog21/lkmpg: https://github.com/sysprog21/lkmpg/issues/122

### DIFF
--- a/.ci/static-analysis.sh
+++ b/.ci/static-analysis.sh
@@ -46,8 +46,9 @@ function do_sparse()
     make sparse || exit 1
     sudo make INST_PROGRAMS=sparse PREFIX=/usr install || exit 1
     popd
+    local SPARSE=$(which sparse)
 
-    make -C examples C=2 2> sparse.log
+    make -C examples C=2 CHECK="$SPARSE" 2> sparse.log
 
     local WARNING_COUNT=$(cat sparse.log | egrep -c " warning:" )
     local ERROR_COUNT=$(cat sparse.log | egrep -c " error:" )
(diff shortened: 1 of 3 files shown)


jserv:
Thank @linD026 for the great work!

PR #126, opened 2021-11-07
7.4: adjust the position of Figure 1.

manbing:
The Figure 1. was located in procfs4.c sample code.
Which make the sample code be fragmented, and hard to read.

Puting the Figure 1. at the same place it occurs in source text.

### DIFF
--- a/lkmpg.tex
+++ b/lkmpg.tex
@@ -1061,7 +1061,7 @@ That means that at the end of function \cpp|stop()|, the function \cpp|start()| 
 This loop finishes when the function \cpp|start()| returns \cpp|NULL|.
 You can see a scheme of this in the Figure~\ref{img:seqfile}.
 
-\begin{figure}
+\begin{figure}[h]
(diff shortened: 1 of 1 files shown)


lkmpg.tex
```diff
@@ -1086,7 +1086,7 @@ \subsection{Manage /proc file with seq\_file}
 The \cpp|seq_file| provides basic functions for \cpp|proc_ops|, such as \cpp|seq_read|, \cpp|seq_lseek|, and some others.
 But nothing to write in the \verb|/proc| file.
 Of course, you can still use the same way as in the previous example.
-
+\newpage
```
jserv, line 1089: I tend to avoid `\newpag` because we would re-organize the chapters later.

manbing: ok, understood. i will remove this, thanks.
please wait.

jserv:
Thank @manbing for contributing!

PR #129, opened 2021-11-10
Update `$(PWD) -> $(shell pwd)` in Makefile in Section 4

mengxinayan:
Hello, everyone

I Update `$(PWD)` to `$(shell pwd)` at Makefile in Section 4. There are 6 changes in total. (Section 4.1, 4.2, 4.6).

My Laptop has installed openSUSE Tumbleweed distro, when I try hello-1 example, it will fail during the make progross. Here are error info:

```shell
╭─mengxinayan@Tumbleweed ~/code/lkmpg
╰─$ sudo make              
make: PWD: No such file or directory
make -C /lib/modules/5.14.14-1-default/build M= modules 
make[1]: Entering directory '/usr/src/linux-5.14.14-1-obj/x86_64/default'
  GEN     Makefile
make[3]: *** No rule to make target 'arch/x86/entry/syscalls/syscall_32.tbl', needed by 'arch/x86/include/generated/uapi/asm/unistd_32.h'.  Stop.
make[2]: *** [/usr/src/linux-5.14.14-1/arch/x86/Makefile:231: archheaders] Error 2
make[1]: *** [/usr/src/linux-5.14.14-1/Makefile:220: __sub-make] Error 2
make[1]: Leaving directory '/usr/src/linux-5.14.14-1-obj/x86_64/default'
make: *** [Makefile:4: all] Error 2
```

It will report [ERROR] because `make: PWD: No such file or directory`.  So I replaced `$(PWD)` with `$(shell pwd)` in Makefile on my laptop, it will work well.

```bash
╭─mengxinayan@Tumbleweed ~/code/lkmpg
╰─$ sudo make 
make -C /lib/modules/5.14.14-1-default/build M=/home/mengxinayan/code/lkmpg modules 
make[1]: Entering directory '/usr/src/linux-5.14.14-1-obj/x86_64/default'
  CC [M]  /home/mengxinayan/code/lkmpg/hello-1.o
  MODPOST /home/mengxinayan/code/lkmpg/Module.symvers
  CC [M]  /home/mengxinayan/code/lkmpg/hello-1.mod.o
  LD [M]  /home/mengxinayan/code/lkmpg/hello-1.ko
  BTF [M] /home/mengxinayan/code/lkmpg/hello-1.ko
Skipping BTF generation for /home/mengxinayan/code/lkmpg/hello-1.ko due to unavailability of vmlinux
make[1]: Leaving directory '/usr/src/linux-5.14.14-1-obj/x86_64/default'
```

So I changed all `$(PWD)` to `$(shell pwd)` in Makefile.

Thanks.

### DIFF
--- a/lkmpg.tex
+++ b/lkmpg.tex
@@ -220,10 +220,10 @@ Now you will need a \verb|Makefile|. If you copy and paste this, change the inde
 obj-m += hello-1.o
 
 all:
-	make -C /lib/modules/$(shell uname -r)/build M=$(PWD) modules
+	make -C /lib/modules/$(shell uname -r)/build M=$(shell pwd) modules
 
 clean:
-	make -C /lib/modules/$(shell uname -r)/build M=$(PWD) clean
+	make -C /lib/modules/$(shell uname -r)/build M=$(shell pwd) clean
 \end{code}
 
 And finally just:
@@ -328,10 +328,10 @@ obj-m += hello-1.o
 obj-m += hello-2.o
 
 all:
-	make -C /lib/modules/$(shell uname -r)/build M=$(PWD) modules
+	make -C /lib/modules/$(shell uname -r)/build M=$(shell pwd) modules
 
 clean:
-	make -C /lib/modules/$(shell uname -r)/build M=$(PWD) clean
+	make -C /lib/modules/$(shell uname -r)/build M=$(shell pwd) clean
 \end{code}
 
 Now have a look at \src{drivers/char/Makefile} for a real world example.
@@ -469,10 +469,10 @@ obj-m += startstop.o
 startstop-objs := start.o stop.o
 
 all:
-	make -C /lib/modules/$(shell uname -r)/build M=$(PWD) modules
+	make -C /lib/modules/$(shell uname -r)/build M=$(shell pwd) modules
 
 clean:
-	make -C /lib/modules/$(shell uname -r)/build M=$(PWD) clean
+	make -C /lib/modules/$(shell uname -r)/build M=$(shell pwd) clean
 \end{code}
 
 This is the complete makefile for all the examples we have seen so far.


linD026:
Hi, mengxinayan.
The reason why action doesn't work is that the `sudo` will fork the process.
man page - `sudo`:
```
   Process model
     There are two distinct ways sudo can run a command.

     If an I/O logging plugin is configured or if the security policy explicitly requests it, a new pseudo-terminal (“pty”) is allocated
     and fork(2) is used to create a second sudo process, referred to as the monitor.  The monitor creates a new terminal session with
     itself as the leader and the pty as its controlling terminal, calls fork(2), sets up the execution environment as described above,
     and then uses the execve(2) system call to run the command in the child process.  The monitor exists to relay job control signals
     between the user's existing terminal and the pty the command is being run in.  This makes it possible to suspend and resume the
     command.  Without the monitor, the command would be in what POSIX terms an “orphaned process group” and it would not receive any
     job control signals from the kernel.  When the command exits or is terminated by a signal, the monitor passes the command's exit
     status to the main sudo process and exits.  After receiving the command's exit status, the main sudo passes the command's exit sta‐
     tus to the security policy's close function and exits.

     If no pty is used, sudo calls fork(2), sets up the execution environment as described above, and uses the execve(2) system call to
     run the command in the child process.  The main sudo process waits until the command has completed, then passes the command's exit
     status to the security policy's close function and exits.  As a special case, if the policy plugin does not define a close func‐
     tion, sudo will execute the command directly instead of calling fork(2) first.  The sudoers policy plugin will only define a close
     function when I/O logging is enabled, a pty is required, or the pam_session or pam_setcred options are enabled.  Note that
     pam_session and pam_setcred are enabled by default on systems using PAM.
```

So, some of the environment variables don't be inherited.

For example, here is the simple Makefile:
```make
all:
	echo $(PWD)
```
```shell
ubuntu@ubuntu:~/temp$ make -p | grep PWD
PWD = /home/ubuntu/temp
OLDPWD = /home/ubuntu
	echo $(PWD)
ubuntu@ubuntu:~/temp$ sudo make -p | grep PWD
	echo $(PWD)
```
It can use the `-E` to preserve them.
```
     -E, --preserve-env
                 Indicates to the security policy that the user wishes to preserve their existing environment variables.  The security
                 policy may return an error if the user does not have permission to preserve the environment.
```
```shell
ubuntu@ubuntu:~/temp$ sudo -E make -p | grep PWD
PWD = /home/ubuntu/temp
OLDPWD = /home/ubuntu
	echo $(PWD)
```
---
Summary, with the `sudo`, I think it is not a bug in the current version. But, I think the same question will come again in the future.  So, maybe it should be changed into shell function or preserve this and writing some annotations.
I prefer the latter, but I will respect the opinions of others.
@jserv , @fennecJ , any ideas?

mengxinayan:
Hi @linD026 

Thanks for your patient explanation. Now I know why the error I mentioned occurs. 

No matter using the shell function or writing annotation, I don't have a preferred solution.

fennecJ:
>Summary, with the sudo, I think it is not a bug in the current version. But, I think the same question will come again in the future. So, maybe it should be changed into shell function or preserve this and writing some annotations.
I prefer the latter...

I preferred the latter, too.

mengxinayan:
The original content is:

>And finally just:
>```console
>make
>```

How about turning it into the following sentence? 

>And finally Just run `make` directly, there is no need to use `sudo make`:
>```console
>make
>```

I want to hear your opinions, thanks.

linD026:
> The original content is:
> 
> 
> 
> >And finally just:
> 
> >```console
> 
> >make
> 
> >```
> 
> 
> 
> How about turning it into the following sentence? 
> 
> 
> 
> >And finally Just run `make` directly, there is no need to use `sudo make`:
> 
> >```console
> 
> >make
> 
> >```
> 
> 
> 
> I want to hear your opinions, thanks.
> 
> 

You should mention the related of sudo make that I talked previously.

mengxinayan:
> 
> And finally just run `make` directly. 
> ```console
> make
> ```
> Note that using `sudo make` might not compile. Because the `sudo` will fork the process. According to the Process Model section in [`sudo` manual](https://man7.org/linux/man-pages/man8/sudo.8.html), some of the environment variables won't be inherited. However, you can use the `-E` flag to preserve them. The command is as follows.
> ```console
> sudo -E make
> ```

Thanks for your advice. How about now?

jserv:
If you check `examples/Makefile` carefully, you will figure the use of `CURDIR`:
```
PWD := $(CURDIR)
```

The evaluation of `CURDIR` is not affected by `sudo`.

mengxinayan:
Hi @linD026 

Thanks for your advice.

> 
> And finally just run make directly.
> ```console
> make
> ```
> Note that using `sudo` make might not compile. Because the `sudo` will fork the process. According to the Process Model section in `sudo` manual, some of the environment variables won't be inherited. However, you can use the `-E` flag to preserve them. The command is as follows.
> ```console
> sudo -E make
> ```
> The following is an example to show that `sudo make` will not inherit the environment variables.
> ```console
> ubuntu@ubuntu:~/temp$ make -p | grep PWD
> PWD = /home/ubuntu/temp
> OLDPWD = /home/ubuntu
> 	echo $(PWD)
> ubuntu@ubuntu:~/temp$ sudo make -p | grep PWD
> 	echo $(PWD)
> ```

But I think if add such an example, it will maybe redundant, because it has some gaps with the topic?  Perhaps, I'm not sure.

mengxinayan:
Hi @jserv 

Thanks for your explanation :).

I will read the manual tomorrow to know more about it.

linD026:
Change to this may be better:

> And finally just run make directly.
> ```console
> make
> ```
> Note that using `sudo make` might not compile. Because the `sudo` will fork the process. According to the Process Model section in sudo manual, some of the environment variables won't be inherited. Here is the `PWD` variable which will be sent into the kernel makefile. For example, here is the simple Makefile:
> ```make
> all:
> 	echo $(PWD)
> ```
> Then, we can use `-p` flag to print out the variable values from the Makefile.
> ```console
> ubuntu@ubuntu:~/temp$ make -p | grep PWD
> PWD = /home/ubuntu/temp
> OLDPWD = /home/ubuntu
> 	echo $(PWD)
> ```
> ```console
> ubuntu@ubuntu:~/temp$ sudo make -p | grep PWD
> 	echo $(PWD)
> ```
> However, you can use the `-E` flag to preserve them. The command is as follows.
> ```console
> ubuntu@ubuntu:~/temp$ sudo -E make -p | grep PWD
> PWD = /home/ubuntu/temp
> OLDPWD = /home/ubuntu
> 	echo $(PWD)
> ```

---

I think the examples are necessary.
Because when the explanation is not clear enough, the same issue may come again.

mengxinayan:
Hi @linD026 

Thank you for the revision and explanation. Your revised content is much clearer than mine. 👍👍👍

I'll create a new PR for it in a minute

mengxinayan:
Hi @linD026 

Here is the new PR #132 . 

Thanks.

PR #132, opened 2021-11-26
Add the explanation about `make` in section 4.1

mengxinayan:
Add the explanation of the difference between `make` and `sudo make` by an example in section 4.1.

### DIFF
--- a/lkmpg.tex
+++ b/lkmpg.tex
@@ -219,6 +219,8 @@ Now you will need a \verb|Makefile|. If you copy and paste this, change the inde
 \begin{code}
 obj-m += hello-1.o
 
+PWD := $(CURDIR)
+
 all:
 	make -C /lib/modules/$(shell uname -r)/build M=$(PWD) modules
 
@@ -231,7 +233,7 @@ And finally just run \verb|make| directly.
 make
 \end{codebash}
 
-Note that using \verb|sudo make| might not compile. Because the \verb|sudo| will fork the process. According to the Process Model section in \verb|sudo| manual, some of the environment variables won't be inherited. Here is the \verb|PWD| variable which will be passed to the kernel Makefile. For example, here is the simple Makefile:
+If there is no \verb|PWD := $(CURDIR)| statement in Makefile, note that using \verb|sudo make| might not compile. Because the \verb|sudo| will fork the process. According to the Process Model section in \verb|sudo| manual, some of the environment variables won't be inherited. Here is the \verb|PWD| variable which will be passed to the kernel Makefile. For example, here is the simple Makefile:
 
 \begin{code}
 all:
@@ -358,6 +360,8 @@ So now we have two real kernel modules under our belt. Adding another module is 
 \begin{code}
 obj-m += hello-1.o
 obj-m += hello-2.o
+
+PWD := $(CURDIR)
 
 all:
 	make -C /lib/modules/$(shell uname -r)/build M=$(PWD) modules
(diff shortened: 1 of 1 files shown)


jserv:
Quoting the GNU Make Manual:
> `CURDIR`
> Set to the absolute pathname of the current working directory.
> 
> For your convenience, when GNU make starts (after it has processed any -C options) it sets the variable CURDIR to the pathname of the current working directory. This value is never touched by make again: in particular note that if you include files from other directories the value of CURDIR does not change. The value has the same precedence it would have if it were set in the makefile (by default, an environment variable CURDIR will not override this value). Note that setting this variable has no impact on the operation of make (it does not cause make to change its working directory, for example).

`CURDIR` is guaranteed to work in Make in `PWD` might be inherited from the shell, the former should be preferred.

linD026 (changes requested):
Looks good.
Now, please add the `CURDIR` part mentioned by jesrv.
Thanks.

mengxinayan:
Hi @jserv @linD026 

Thanks. I want to replace all `PWD` with `CURDIR` in the book, and explain the situation mentioned above when using `PWD`.  Is it a good idea?

linD026:
> Hi @jserv @linD026
> 
> Thanks. I want to replace all `PWD` with `CURDIR` in the book, and explain the situation mentioned above when using `PWD`. Is it a good idea?

I prefer to add the following statement which is from `example/Makefile` instead of replacing the `PWD` with `CURDIR`.
```make
PWD := $(CURDIR)
```

mengxinayan:
Hi @linD026 

Thanks for your suggestion. I will revise it by following your advice.

lkmpg.tex
```diff
@@ -219,18 +219,52 @@ \subsection{The Simplest Module}
...
 	make -C /lib/modules/$(shell uname -r)/build M=$(PWD) clean
 \end{code}
 
-And finally just:
+And finally just run \verb|make| directly.
 \begin{codebash}
 make
 \end{codebash}
 
+If there is no \verb|PWD := $(CURDIR)| statement in Makefile, note that using \verb|sudo make| might not compile. Because the \verb|sudo| will fork the process. According to the Process Model section in \verb|sudo| manual, some of the environment variables won't be inherited. Here is the \verb|PWD| variable which will be passed to the kernel Makefile. For example, here is the simple Makefile:
```
jserv, line 236: Add a hyperlink to `sudo` manual.

mengxinayan: Thanks, I have added it.

jserv:
@mengxinayan, I wonder the reason and/or motivation why you close this pull request. I didn't find anything wrong.

lkmpg.tex
```diff
@@ -327,14 +361,16 @@ \subsection{Hello and Goodbye}
...
+
 all:
 	make -C /lib/modules/$(shell uname -r)/build M=$(PWD) modules
 
 clean:
 	make -C /lib/modules/$(shell uname -r)/build M=$(PWD) clean
 \end{code}
 
-Now have a look at \src{drivers/char/Makefile} for a real world example.
+Now have a look at {drivers/char/Makefile} for a real world example.
```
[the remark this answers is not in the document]
mengxinayan: Sorry, this is an accidental error modification by me :(

mengxinayan:
Hi @jserv @linD026 

Thanks for your suggestions. I open a new PR #133  to integrate all modifications, making it simply.

mengxinayan:
> @mengxinayan, I wonder the reason and/or motivation why you close this pull request. I didn't find anything wrong.

I just want to make all modifications simply. Maybe deleting it is not a good idea ? I have no idea about this.

PR #135, opened 2021-12-22
Introduce the new method for chardev registration

RinHizakura:
When I am using this book, I found that the book seems only introduced the char device registration using the older `register_chrdev` / `unregister_chrdev` API. I think we should introduce the cdev interface which is newer and would be better for new driver programmers in most cases. This commit is trying to give the brief overview and also provide comparable codes between two different interfaces.

There may be many insufficient and looking for discussion to change/improve it: 
1. I'm not sure the sentences in the book are enough to introduce the concept, or I should give more explanation.
2. The codes for the old and the new are separated into different files. This produces duplicate codes although make me easy to do those ci(static analysis and build) test. I'm not sure if you would like an alternative approach which is using something like `#define NEW_API` to merge two files.

### DIFF
--- a/lkmpg.tex
+++ b/lkmpg.tex
@@ -896,6 +896,43 @@ First, the driver itself can print the newly assigned number and we can make the
 Second, the newly registered device will have an entry in \verb|/proc/devices|, and we can either make the device file by hand or write a shell script to read the file in and make the device file.
 The third method is that we can have our driver make the device file using the \cpp|device_create| function after a successful registration and \cpp|device_destroy| during the call to \cpp|cleanup_module|.
 
+However, \cpp|register_chrdev()| would occupy a range of minor numbers associated with the given major.
+The recommended way to reduce waste for char device registration is using cdev interface.
+
+The newer interface completes the char device registration in two distinct steps.
+First, we should register a range of device numbers, which can be completed with \cpp|register_chrdev_region| or \cpp|alloc_chrdev_region|.
+
+\begin{code}
+int register_chrdev_region(dev_t from, unsigned count, const char *name);
+int alloc_chrdev_region(dev_t *dev, unsigned baseminor, unsigned count, const char *name);
+\end{code}
+
+The choose of two different functions depend on whether you know the major numbers for your device.
+Using \cpp|register_chrdev_region| if you know the device major number and \cpp|alloc_chrdev_region| if you would like to allocate a dynamicly-allocated major number.
+
+Second, we should initialize the data structure \cpp|struct cdev| for our char device and associate it with the device numbers.
+To initialize the \cpp|struct cdev|, we can achieve by the similar sequence of the following codes.
+
+\begin{code}
+struct cdev *my_dev = cdev_alloc();
+my_cdev->ops = &my_fops;
+\end{code}
+
+However, the common usage pattern will embed the \cpp|struct cdev| within a device-specific structure of your own.
+In this case, we'll need \cpp|cdev_init| for the initialization.
+
+\begin{code}
+void cdev_init(struct cdev *cdev, const struct file_operations *fops);
+\end{code}
+
+Once we finish the initialization, we can add the char device to the system by using the \cpp|cdev_add|.
+
+\begin{code}
+int cdev_add(struct cdev *p, dev_t dev, unsigned count);
+\end{code}
+
+To find a example using the interface, you can see \verb|ioctl.c| described in section \ref{sec:device_files}.
+
 \subsection{Unregistering A Device}
 \label{sec:unregister_device}
 We can not allow the kernel module to be \sh|rmmod|'ed whenever root feels like it.


lkmpg.tex
```diff
@@ -944,6 +982,11 @@ \subsection{chardev.c}
 
 \samplec{examples/chardev.c}
 
+The following codes shows the rewriting of chardev.c using cdev interface. You can compare
```
jserv, line 985: Remove "codes"

jserv: Always keep "one sentence one line" policy. That is, you should split the sentence starting with "You can compare" into the next line.

lkmpg.tex
```diff
@@ -896,6 +896,41 @@ \subsection{Registering A Device}
...
+
+Second, we should initialize the cdev structure and associate it with the deivce numbers.
+To initialize a cdev structure, we can achieve by the similar sequence of the following codes.
+
+\begin{code}
+struct cdev *my_dev = cdev_alloc();
+my_cdev->ops = &my_fops;
+\end{code}
+
+However, the common usage pattern will embed the cdev structure within a device-specific structure of your own.
```
jserv, line 921: Use proper annotation for `cdev`. That is, `\cpp|struct cdev|` instead of `cdev structure`.

linD026 (changes requested):
Hi, The new API already exists in the `ioctl.c`.
```cpp
static int ioctl_init(void)
{
    dev_t dev;
    int alloc_ret = 0;
    int cdev_ret = 0;
    alloc_ret = alloc_chrdev_region(&dev, 0, num_of_dev, DRIVER_NAME);

    if (alloc_ret)
        goto error;

    test_ioctl_major = MAJOR(dev);
    cdev_init(&test_ioctl_cdev, &fops);
    cdev_ret = cdev_add(&test_ioctl_cdev, dev, num_of_dev);

    if (cdev_ret)
        goto error;

    pr_alert("%s driver(major: %d) installed.\n", DRIVER_NAME,
             test_ioctl_major);
    return 0;
error:
    if (cdev_ret == 0)
        cdev_del(&test_ioctl_cdev);
    if (alloc_ret == 0)
        unregister_chrdev_region(dev, num_of_dev);
    return -1;
}
```
But it is still happy to see those APIs have an explanation.
You can keep the explanation in `chardev` section and mention where the new API be in use.
Also, add the description for using the new API in `ioctl` section.

RinHizakura:
> You can keep the explanation in `chardev` section and mention where the new API be in use. Also, add the description for using the new API in `ioctl` section.

Does it means we are going to remove  `chrdev_new.c`  and leave the cdev interface example in `ioctl.c` ?

linD026:
> Does it mean we are going to remove  `chrdev_new.c` (which is added in this PR) and leave the cdev interface example in `ioctl.c`  by mentioning it in `chardev` section?
> 

Yes.

lkmpg.tex
```diff
@@ -896,6 +896,43 @@ \subsection{Registering A Device}
...
+void cdev_init(struct cdev *cdev, const struct file_operations *fops);
+\end{code}
+
+Once we finish the initialization, we can add the char device to the system by using the \cpp|cdev_add|.
+
+\begin{code}
+int cdev_add(struct cdev *p, dev_t dev, unsigned count);
+\end{code}
+
+To find a example using the interface, you can see ioctl.c which is located in section \ref{sec:device_files}.
```
linD026, line 934: To make it consistent with another use case ( `\verb|hello-1.c|` ), use `\verb|ioctl.c|` instead of `ioctl.c`.
Except for this tiny problem and previously Jserv requested, others look good to me.

linD026 (changes requested):

lkmpg.tex
```diff
@@ -896,6 +896,43 @@ \subsection{Registering A Device}
...
+
+\begin{code}
+int register_chrdev_region(dev_t from, unsigned count, const char *name);
+int alloc_chrdev_region(dev_t *dev, unsigned baseminor, unsigned count, const char *name);
+\end{code}
+
+The choose of two different functions depend on whether you know the major numbers for your device.
+Using \cpp|register_chrdev_region| if you know the device major number and \cpp|alloc_chrdev_region| if you would like to allocate a dynamicly-allocated major number.
+
+Second, we should initialize the data structure for our char device, which is \cpp|struct cdev|, and associate it with the device numbers.
```
jserv, line 913: Remove "which is" for simplicity.

lkmpg.tex
```diff
@@ -896,6 +896,43 @@ \subsection{Registering A Device}
...
+void cdev_init(struct cdev *cdev, const struct file_operations *fops);
+\end{code}
+
+Once we finish the initialization, we can add the char device to the system by using the \cpp|cdev_add|.
+
+\begin{code}
+int cdev_add(struct cdev *p, dev_t dev, unsigned count);
+\end{code}
+
+To find a example using the interface, you can see \verb|ioctl.c| which is located in section \ref{sec:device_files}.
```
jserv, line 934: Change to
> see \verb|ioctl.c| described in section \ref{sec:device_files}.

lkmpg.tex
```diff
@@ -896,6 +896,43 @@ \subsection{Registering A Device}
 Second, the newly registered device will have an entry in \verb|/proc/devices|, and we can either make the device file by hand or write a shell script to read the file in and make the device file.
 The third method is that we can have our driver make the device file using the \cpp|device_create| function after a successful registration and \cpp|device_destroy| during the call to \cpp|cleanup_module|.
 
+However, \cpp|register_chrdev()| would occupy a range of minor numbers associated with the given major.
+A more recommended way to register char device is using cdev interface.
```
jserv, line 900: Change "A more recommended way" to "The recommended way to reduce waste"

RinHizakura:
All the mentioned requirements are changed accordingly, thanks!

jserv:
Thank @RinHizakura for contributing!

PR #136, opened 2021-12-30
Add book cover

jkrvivian:
Insert image from #4 as the book cover, and add authors with tikz.


jserv:
Thank @jkrvivian for the effort! However, we have to ensure the HTML generation as well. Can you run `make html` and figure out anything wrong?

lkmpg.tex
```diff
@@ -44,10 +45,20 @@
 \title{The Linux Kernel Module Programming Guide}
 \begin{document}
 
+
```
jserv, line 48: Minimize the necessary changes. Don't add unexpected two blank lines.

lkmpg.tex
```diff
@@ -44,10 +45,20 @@
...
+
 \maketitle
 \ifdefined\HCode
+\includegraphics{assets/cover-with-names.png}
 % turn off TOC
 \else
+\pagestyle{empty}
+\begin{tikzpicture}[remember picture, overlay]
+		\node at (current page.center) {\includegraphics[width=\paperwidth, height=\paperheight]{assets/cover.png}}; \\
```
jserv, line 57: Use two spaces for indention.

jserv:
Thank @jkrvivian for contributing!

PR #137, opened 2022-02-12
Fix a logic error in examples/ioctl.c

NOVBobLee:
Change the `alloc_ret` and `cdev_ret` initial values to non-zero, or it
will `cdev_del()` uninitialized and not-yet-added `test_ioctl_cdev` after
`alloc_chrdev_region()` failed.

### DIFF
--- a/examples/ioctl.c
+++ b/examples/ioctl.c
@@ -149,8 +149,8 @@ static struct file_operations fops = {
 static int ioctl_init(void)
 {
     dev_t dev;
-    int alloc_ret = 0;
-    int cdev_ret = 0;
+    int alloc_ret = -1;
+    int cdev_ret = -1;
     alloc_ret = alloc_chrdev_region(&dev, 0, num_of_dev, DRIVER_NAME);
 
     if (alloc_ret)


linD026 (approved):
Hi NOVBobLee, 
I'm sorry for a long time with no response. I just see it right now.

According to the source code, `alloc_chrdev_region()` will return a negative integer when failed. Otherwise, it will return zero.
```cpp
225 /**                                                                              
226  * alloc_chrdev_region() - register a range of char device numbers               
227  * @dev: output parameter for first assigned number                              
228  * @baseminor: first of the requested range of minor numbers                     
229  * @count: the number of minor numbers required                                  
230  * @name: the name of the associated device or driver                            
231  *                                                                               
232  * Allocates a range of char device numbers.  The major number will be           
233  * chosen dynamically, and returned (along with the first minor number)          
234  * in @dev.  Returns zero or a negative error code.                              
235  */                                                                              
236 int alloc_chrdev_region(dev_t *dev, unsigned baseminor, unsigned count,          
237                         const char *name)                                        
238 {                                                                                
239         struct char_device_struct *cd;                                           
240         cd = __register_chrdev_region(0, baseminor, count, name);                                    
241         if (IS_ERR(cd))                                                          
242                 return PTR_ERR(cd);                                              
243         *dev = MKDEV(cd->major, cd->baseminor);                                  
244         return 0;                                                                
245 }  
```
If the function fails, the if statement will be true (no-zero value, which is a negative integer), it will execute the goto statement then jump to the error code.
The error code will check the return value where did it failed and deal with it properly.
But it compares the return value with zero for checking the return value.
It is a logical error when the initial value is zero.

So, YES there are some logical errors in the error code.
Thanks!

jserv:
Thank @linD026 for reviewing. I think @NOVBobLee can improve the git commit messages by analyzing the relevant code snip as @linD026 attempted to discuss.

NOVBobLee:
Thanks for reviewing!
I will modify the commit message more readable.

jserv:

Don't use markdown syntax in the body of git commit messages. Instead, always use plaintext.

PR #140, opened 2022-03-07
Fix typo, grammar and remove duplicated words

steven1lung:
Replaced 'a' with 'an' since a vowel sound is after.


jserv:
Thank @steven1lung for contributing!

PR #142, opened 2022-03-16
Fix comment mistake

SmallHanley:


linD026:
Hi @SmallHanley ,
The kprobe/parameter methods are used at v5.7+.
```cpp
- 5.7 <= ver : Kprobes or specific kernel module parameter
```
So I think this is not a mistake.

jserv:
> Hi @SmallHanley , The kprobe/parameter methods are used at v5.7+.
> So I think this is not a mistake.

However, the notation was not intuitive. Instead, we can say, "Linux v5.7+"

SmallHanley:
Hi @linD026 ,
I see. I mistook `5.7 <= ver` for `ver <= 5.7`.
The original one is correct.

PR #145, opened 2022-04-05
Introduce Virtual Input Device Driver

linD026:
Add the new section of device driver, vinput[1].

[1] https://github.com/sysprog21/vinput

### DIFF
--- a/examples/vinput.c
+++ b/examples/vinput.c
@@ -1,0 +1,398 @@
+#include <linux/cdev.h>
+#include <linux/input.h>
+#include <linux/module.h>
+#include <linux/slab.h>
+#include <linux/spinlock.h>
+
+#include <asm/uaccess.h>
+
+#include "vinput.h"
+
+#define DRIVER_NAME "vinput"
+
+#define dev_to_vinput(dev) container_of(dev, struct vinput, dev)
+
+static DECLARE_BITMAP(vinput_ids, VINPUT_MINORS);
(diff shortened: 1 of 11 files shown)


lkmpg.tex
```diff
@@ -1857,6 +1857,81 @@ \subsection{Symmetric key encryption}
 
 \samplec{examples/cryptosk.c}
 
+\section{Virtual Input Device Driver}
+\label{sec:vinput}
+vinput is a API to allow easy development of virtual input drivers.
```
jserv, line 1862: The motivation of `vinput` is to complement the documentation [Linux Input Subsystem kernel API](https://www.kernel.org/doc/html/latest/input/input_kapi.html), which lacks of complete examples. You can describe some essential concepts from the kernel docs.

examples/vinput.c
```diff
@@ -0,0 +1,399 @@
...
+
+    unregister_chrdev(vinput_dev, DRIVER_NAME);
+    class_unregister(&vinput_class);
+}
+
+module_init(vinput_init);
+module_exit(vinput_end);
+
+MODULE_LICENSE("GPL");
+MODULE_AUTHOR("Tristan Lelong <tristan.lelong@blunderer.org>");
```
jserv, line 398: Remove the name from the source code, and then you should mention Tristan Lelong in the section "Acknowledgements."

PR #154, opened 2022-06-25
module_init and module_exit are defined in module.h

leovincentseles:
Both module_init and module_exit are defined in
include/linux/module.h


jserv:
Thank @leovincentseles for contributing!

PR #156, opened 2022-07-02
Adjust "in the mean time" description of tasklet

### LINKED ISSUE #152: [14.1 Tasklets] init function may be interrupted by tasklet
In [14.1 Tasklets](https://sysprog21.github.io/lkmpg/#tasklets):

> Here is an example tasklet module. The tasklet_fn function runs for a few seconds and **in the mean time execution of the example_tasklet_init function continues to the exit point.**

```c
static void tasklet_fn(unsigned long data) 
{ 
    pr_info("Example tasklet starts\n"); 
    mdelay(5000); 
    pr_info("Example tasklet ends\n"); 
} 
 
static DECLARE_TASKLET_OLD(mytask, tasklet_fn); 
 
static int example_tasklet_init(void) 
{ 
    pr_info("tasklet example init\n"); 
    tasklet_schedule(&mytask); 
    mdelay(200); 
    pr_info("Example tasklet init continues...\n"); 
    return 0; 
} 
```

> So with this example loaded dmesg should show:
> 
> tasklet example init
> Example tasklet starts
> Example tasklet init continues...
> Example tasklet ends

However, when `example_tasklet_init` is in `mdelay(200)`, the thread may be interrupted by the `tasklet_fn` since it's a softirq, thus the output becomes:

```bash
qiuhao@pc:~/tmp/lkmpg/examples$ sudo insmod example_tasklet.ko
qiuhao@pc:~/tmp/lkmpg/examples$ sudo dmesg | tail -n 4
[1293636.040724] tasklet example init
[1293636.043172] Example tasklet starts
[1293641.003428] Example tasklet ends
[1293641.198868] Example tasklet init continues...
```

Maybe we should fix the description "in the mean time..."?

linD026:
Since the init function may be interrupted. Tweak the description of
"in the mean time".

Close #152

### DIFF
--- a/lkmpg.tex
+++ b/lkmpg.tex
@@ -1763,7 +1763,8 @@ For example, when triggered from an interrupt, whereas work queues are more comp
 \subsection{Tasklets}
 \label{sec:tasklet}
 Here is an example tasklet module.
-The \cpp|tasklet_fn| function runs for a few seconds and in the mean time execution of the \cpp|example_tasklet_init| function continues to the exit point.
+The \cpp|tasklet_fn| function runs for a few seconds.
+In the meantime, execution of the \cpp|example_tasklet_init| function may continue to the exit point, depending on whether it is interrupted by \textbf{softirq}.
 
 \samplec{examples/example_tasklet.c}
 


lkmpg.tex
```diff
@@ -1763,7 +1763,8 @@ \section{Scheduling Tasks}
 \subsection{Tasklets}
 \label{sec:tasklet}
 Here is an example tasklet module.
-The \cpp|tasklet_fn| function runs for a few seconds and in the mean time execution of the \cpp|example_tasklet_init| function continues to the exit point.
+The \cpp|tasklet_fn| function runs for a few seconds.
+In the meantime, execution of the \cpp|example_tasklet_init| function may continue to the exit point, which depending on whether it is interrupted by \textbf{softirq}.
```
jserv, line 1767: Replace "which depending on" with "depending on"

Issue #158, opened 2022-08-27
Refine doc & code


minghu6:
1. Move the confusing note about `struct proc_ops` replacing with `struct file_operations` on linux >=5.6:
https://github.com/sysprog21/lkmpg/blob/637e707a1ab7efa62234a374c62ffb040800aca8/lkmpg.tex#L957-L990
The `struct proc_ops` would replace the `struct file-operations` only when create proc  releated defined on <linux/proc_fs.h>
In other word, regist device doesn't be affected, still using `struct file_operations`. It's confusing write this note above the regist device code (FOE ME, I HAVE SPENT PLENTY OF TIME ON SEARCHING HOW TO REGIST DEVICE USING STRUCT PROC_OPS)
I highly recommend to move this note down to next code example which replacing takes place truely!

2.  https://github.com/sysprog21/lkmpg/blob/637e707a1ab7efa62234a374c62ffb040800aca8/examples/chardev.c#L25-L39
The msg char array should be declared with length of `BUF_LEN + 1` since `BUF_LEN` is max length of message.

3. https://github.com/sysprog21/lkmpg/blob/637e707a1ab7efa62234a374c62ffb040800aca8/examples/chardev.c#L152
On `device_write`, It's prefered to use `EOPNOTSUPP` instead of `EINVAL` since the former raises unsupported error, the latter raises invalid argument.

4. Improve read/write fucntion code among `fsproc{1-3}.c`:
https://github.com/sysprog21/lkmpg/blob/637e707a1ab7efa62234a374c62ffb040800aca8/examples/procfs3.c#L23-L40
https://github.com/sysprog21/lkmpg/blob/637e707a1ab7efa62234a374c62ffb040800aca8/examples/procfs3.c#L41-L53

These code are a little wired and far from decent, the parameter `loff_t *` is ignored incorrectly and so the behavior is quite different from normal read/write.I've reimplemented read/write code, here is full code:

```c
#include <linux/kernel.h>
#include <linux/module.h>
#include <linux/proc_fs.h>
#include <linux/uaccess.h>
#include <linux/version.h>
// #include <linux/slab.h>  // kmalloc etc.

/* Header */
#define do_sslice(news, s, n) for(int i=0;i<n;++i) { news[i]=s[i]; };news[n]='\0';

#define proc_name "hello"
#define PROC_BUF_CAP (size_t)1024

static ssize_t proc_read(struct file *filp, char __user *buffer, size_t size,
                         loff_t *offset);
[...]

PR #163, opened 2022-09-07
procfs{2, 3}: Change to use offset parameter

linD026:
To make sure the behavior of the read and write operations are correct with offset, update it each time. Also, since it is using the offset, modify the part of read for removing unnecessary variable.

---
Related discussion:
* https://github.com/sysprog21/lkmpg/issues/158

### DIFF
--- a/examples/procfs3.c
+++ b/examples/procfs3.c
@@ -8,12 +8,13 @@
 #include <linux/sched.h>
 #include <linux/uaccess.h>
 #include <linux/version.h>
+#include <linux/minmax.h>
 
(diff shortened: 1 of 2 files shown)


examples/procfs3.c
```diff
@@ -23,17 +23,16 @@ static unsigned long procfs_buffer_size = 0;
...
+    if (*offset || procfs_buffer_size == 0) {
         pr_debug("procfs_read: END\n");
-        finished = 0;
+        *offset = 0;
         return 0;
     }
-    finished = 1;
-
+    if (length < procfs_buffer_size)
+        procfs_buffer_size = length;
```
jserv, line 32: Can you make use of macro `min`?

PR #174, opened 2022-10-24
fixed typo (line 1019)

jcallemc:
Changed "The choose ..." to "The choice ..."

### DIFF
--- a/lkmpg.tex
+++ b/lkmpg.tex
@@ -1016,7 +1016,7 @@ int register_chrdev_region(dev_t from, unsigned count, const char *name);
 int alloc_chrdev_region(dev_t *dev, unsigned baseminor, unsigned count, const char *name);
 \end{code}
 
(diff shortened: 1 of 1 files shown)


lkmpg.tex
```diff
@@ -1016,7 +1016,7 @@ \subsection{Registering A Device}
 int alloc_chrdev_region(dev_t *dev, unsigned baseminor, unsigned count, const char *name);
 \end{code}
 
-The choose of two different functions depend on whether you know the major numbers for your device.
+The choice of two different functions depend on whether you know the major numbers for your device.
```
jserv, line 1019: Should we use "depends on" ?

jcallemc: Ah, of course. Maybe even "The choice between two different functions depends on ..."?

jserv: Sounds good! Let's rephrase.

PR #180, opened 2022-11-21
Draft: Introduce Ftrace in the syscall chapter

### LINKED ISSUE #175: Any chance to introduce Ftrace in the syscall chapter (or elsewhere) in this book?
In the syscall chapter, syscall hooking is introduced, so I think it may be a good opportunity to go a step deeper to talk about Ftrace and hooking, and maybe kernel live patching.

I can provide some introductory text and sample code for syscall hooking if anyone is interested.

nickchen120235:
Closes #175

- [ ] Write the syscall stealing example using Ftrace
- [ ] Add contents into the book


examples/syscall-ftrace.c
```diff
@@ -0,0 +1,175 @@
...
+#include <linux/slab.h>
+/** This is what we're using here. */
+#include <linux/ftrace.h>
+
+MODULE_LICENSE("GPL");
+
+#define MAX_FILENAME_SIZE 200
+
+/* UID we want to spy on - will be filled from the command line. */
+static int uid;
```
linD026, line 25: So, do we have to give the input value every time when running this example?
If so, we should not execute this example during the checking. Then, it will be added to the non-working list (located at  `/lkmpg/.ci/non-working`).

Also, `uid` should be initialized.

Thanks.

linD026 (changes requested):
I suggest we should also provide a user script to work with this example.
It can be located at `/lkmpg/example/other` directory.

examples/syscall-ftrace.c
nickchen120235: I followed the logic from [`syscall.c`](https://github.com/sysprog21/lkmpg/blob/9c25e59b439fcdd5fb473c9635716a1de03853b6/examples/syscall.c#L65-L66) so yes `uid` is given at `insmod` as an parameter.

linD026: Hmm, I looked at the syscall.c again. In the syscall.c, the uid is unused variable since we removed the if-statement (check the current->uid is same as the uid) in our_sys_open(). (Original version: https://tldp.org/LDP/lkmpg/2.6/html/x978.html)

It's weird, I will fix it.

However, the uid is only used at the self-defined syscall (the use case of printk() is unimportant), which means we don't care about it when doing the checking (insert and remove the module several times, so it will test the registration only).
So, we don't have to add the syscall.c to the non-working list.
And, I strongly believe the  complicated registration of syscall.c should be tested in the checking.

Back to the PR, I roughly look at the code. Seems like it is similar to the syscall.c. So, I suggest the code here should also add the if-statement in our_sys_openat(), just like the original our_sys_open().
And, we should test the registration too.

BTW, for the initialization of uid, I just forgot the static variable will be initialized by zero. So, it's fine.

nickchen120235: Yeah I added the missing comparison [here](https://github.com/sysprog21/lkmpg/pull/180/commits/9b7be185b1c9dab0654bc9932bcb8e2ba871b28a).

And for `uid` initialization, I explicitly set it to zero but it should work anyways.

nickchen120235: Also `syscall.c` doesn't come with a user-space program in `example/other`, so I assume I don't need one here?

linD026: Yeah, we can leave it right now.
I will consider again do we need the user programs or not.

Thanks.

examples/syscall-ftrace.c
```diff
@@ -0,0 +1,222 @@
...
+typedef struct ftrace_hook {
+    unsigned long nr; // syscall name
+    void *new; // hook function
+    void *orig; // original function
+
+    unsigned long address; // address to the original function
+    struct ftrace_ops ops; // ftrace structure
+} ftrace_hook_t;
+
+// clang-format off
```
jserv, line 51: Remove `clang-format off` annotation because it would appear in the content.

examples/syscall-ftrace.c
```diff
@@ -0,0 +1,222 @@
...
+#include <linux/init.h>
+#include <linux/module.h>
+#include <linux/moduleparam.h>
+#include <linux/version.h>
+#include <linux/unistd.h>
+#include <linux/kprobes.h>
+#include <linux/sched.h>
+#include <linux/uaccess.h>
+#include <linux/slab.h>
+/** This is what we're using here. */
```
jserv, line 23: Remove this comment. Always minimize sample files.

lkmpg.tex
```diff
@@ -1554,6 +1554,97 @@ \section{System Calls}
...
+If the call is from the kernel, our auditing function is called.
+Otherwise, nothing happens.
+The check is important because we're only ``decorating'' the original syscall.
+Our auditing function contains call to the original \cpp|sys_execve|,
+which will trigger the callback function again.
+It'll be an infinite loop if there's no check performed.
+
+\begin{figure}[h]
+  \centering
+  \includegraphics[width=\textwidth]{assets/syscall/flow.jpg}
```
jserv, line 1604: Utilize TikZ for drawing. See https://texample.net/tikz/examples/pgf-umlsd/
Avoid putting bitmap files.

examples/syscall-ftrace.c
```diff
@@ -0,0 +1,222 @@
...
+    struct ftrace_ops ops; // ftrace structure
+} ftrace_hook_t;
+
+// clang-format off
+#define PREPARE_HOOK(_nr, _hook, _orig)              \
+    {                                                \
+        .nr = (_nr), .new = (_hook), .orig = (_orig) \
+    }
+
+unsigned long **sys_call_table;
```
linD026, line 57: Should be static type.

examples/syscall-ftrace.c
```diff
@@ -0,0 +1,222 @@
...
+                                 struct ftrace_ops *ops, struct pt_regs *regs)
+{
+    ftrace_hook_t *hook = container_of(ops, ftrace_hook_t, ops);
+    if (!within_module(parent_ip, THIS_MODULE))
+        regs->ip = (unsigned long)hook->new;
+}
+
+#endif /** Version >= v5.11 */
+
+int install_hook(ftrace_hook_t *hook)
```
linD026, line 130: Should be static function.

examples/syscall-ftrace.c
```diff
@@ -0,0 +1,222 @@
...
+    err = register_ftrace_function(&hook->ops);
+    if (err) {
+        pr_err("[syscall-ftrace] register_ftrace_function() failed: %d\n", err);
+        return err;
+    }
+
+    return 0;
+}
+
+void remove_hook(ftrace_hook_t *hook)
```
linD026, line 157: Should be static function.

examples/syscall-ftrace.c
```diff
@@ -0,0 +1,222 @@
...
+        return original_call(regs);
+
+    /**
+     * This may only work in x86_64 because getting parameters
+     * from CPU registers is architecture-dependent.
+     * 
+     * Change regs->si to appropriate registers
+     * if you are trying on different architecture.
+     */
+    // clang-format off
```
linD026, line 189: Why clang-format off here?

examples/syscall-ftrace.c
```diff
@@ -0,0 +1,222 @@
...
+static int resolve_address(ftrace_hook_t *hook)
+{
+    static struct kprobe kp = { .symbol_name = "kallsyms_lookup_name" };
+    unsigned long (*kallsyms_lookup_name)(const char *name);
+    register_kprobe(&kp);
+    kallsyms_lookup_name = (unsigned long (*)(const char *))kp.addr;
+    unregister_kprobe(&kp);
+
+    if (kallsyms_lookup_name)
+        pr_info("[syscall-ftrace] kallsyms_lookup_name is found at 0x%lx\n",
```
linD026, line 74: For `[syscall-ftrace]` prefix, you can use the following macro to make the code cleaner.
```cpp
#undef pr_fmt
#define pr_fmt(fmt) "[syscall-ftrace] " fmt
```

linD026 (changes requested):

linD026 (changes requested):
Here is the warning, please fix it too.
```
/home/runner/work/lkmpg/lkmpg/examples/syscall-ftrace.c:190:9: warning: ignoring return value of ‘copy_from_user’, declared with attribute warn_unused_result [-Wunused-result]
  190 |     if (copy_from_user(kfilename, (char __user *)regs->si, MAX_FILENAME_SIZE) < 0) {
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

nickchen120235:
[`syscall-ftrace.c`](https://github.com/nickchen120235/lkmpg/commit/6821a2bdcb6355ea04622f1b79a2040186a8dc03) should be good to go.

lkmpg.tex
```diff
@@ -1554,6 +1554,96 @@ \section{System Calls}
...
+                   struct ftrace_ops *ops, struct pt_regs *regs);
+\end{code}
+
+Above is the basic structure of the callback function, where the parameters are
+
+\begin{itemize}
+  \item \cpp|ip|: The instruction pointer of the function being traced.
+  \item \cpp|parent_ip|: The instruction pointer of the caller of the traced function.
+  \item \cpp|ops|: A pointer to \cpp|ftrace_ops| that was used to register the callback.
+  \item \cpp|regs|: If \cpp|FTRACE_OPS_FL_SAVE_REGS| or \cpp|FTRACE_OPS_FL_SAVE_REGS_IF_SUPPORTED| are set in the \cpp|ftrace_ops| structure, then this will be pointing to the \cpp|pt_regs| structure like it would be if an breakpoint was placed at the start of the function where \verb|ftrace| was tracing for CPU register access. Otherwise it either contains garbage, or \cpp|NULL|. Do notice that in kernel version later than v5.11, this is replaced with \cpp|struct ftrace_regs *fregs|, with the original \cpp|pt_regs| accessible by \cpp|fregs->regs|.
```
jserv, line 1578: There should be a newline after each sentence.

examples/syscall-ftrace.c
```diff
@@ -0,0 +1,222 @@
...
+ */
+static int resolve_address(ftrace_hook_t *hook)
+{
+    static struct kprobe kp = { .symbol_name = "kallsyms_lookup_name" };
+    unsigned long (*kallsyms_lookup_name)(const char *name);
+    register_kprobe(&kp);
+    kallsyms_lookup_name = (unsigned long (*)(const char *))kp.addr;
+    unregister_kprobe(&kp);
+
+    if (kallsyms_lookup_name)
```
jserv, line 73: Rewrite as following:
```c
if (!kallsyms_lookup_name) {
    pr_err("kallsyms_lookup_name is not found!\n");
    return -1;
}
pr_info("kallsyms_lookup_name is found at 0x%lx\n",
        (unsigned long)kallsyms_lookup_name);

```

examples/syscall-ftrace.c
```diff
@@ -0,0 +1,222 @@
...
+        pr_info("kallsyms_lookup_name is found at 0x%lx\n",
+                (unsigned long)kallsyms_lookup_name);
+    else {
+        pr_err("kallsyms_lookup_name is not found!\n");
+        return -1;
+    }
+
+    sys_call_table = (unsigned long **)kallsyms_lookup_name("sys_call_table");
+    if (sys_call_table)
```
jserv, line 82: Ditto. Shorten the code.

lkmpg.tex
```diff
@@ -1554,6 +1554,99 @@ \section{System Calls}
...
+and the callback function (\cpp|ftrace_thunk|).
+We need both \cpp|FTRACE_OPS_FL_SAVE_REGS| and \cpp|FTRACE_OPS_FL_IPMODIFY| because we're modifying \cpp|ip|.
+Inside \cpp|ftrace_thunk| is what the magic happens.
+We check if it is called from within the module,
+if not then it modifies the instruction pointer to our ``spying'' function.
+The check is performed by checking whether \cpp|parent_ip| is within this module.
+During the first call, \cpp|parent_ip| points to somewhere within the kernel,
+while during the second call it points to somewhere in our ``spying'' function, which is within the module.
+
+\samplec{examples/syscall-ftrace.c}
```
jserv, line 1648: Since `syscall-ftrace` accepts one parameter, can you show an example for usage?

lkmpg.tex
nickchen120235: Also, I have to rotate the figure by 90 degrees to minimize the overflow ~~although it overflows anyway~~. Will it affect the output of the website or should I rotate it back and let it overflow?

jserv: Alternatively, you can just render the partial sequences.

examples/syscall-ftrace.c
```diff
@@ -0,0 +1,220 @@
...
+
+/**
+ * For the sake of simplicity, only the kprobe method is included.
+ * If you want to know more about different methods to get
+ * kallsyms_lookup_name, see syscall.c.
+ */
+static int resolve_address(ftrace_hook_t *hook)
+{
+    static struct kprobe kp = { .symbol_name = "kallsyms_lookup_name" };
+    unsigned long (*kallsyms_lookup_name)(const char *name);
```
linD026, line 68: Always keep one blank line between the variable declaration and the following statements.
This is not actually described in the coding style guide, however, most kernel codes follow this style.

examples/syscall-ftrace.c
```diff
@@ -0,0 +1,220 @@
...
+static void notrace ftrace_thunk(unsigned long ip, unsigned long parent_ip,
+                                 struct ftrace_ops *ops,
+                                 struct ftrace_regs *fregs)
+{
+    ftrace_hook_t *hook = container_of(ops, ftrace_hook_t, ops);
+    if (!within_module(parent_ip, THIS_MODULE))
+        fregs->regs.ip = (unsigned long)hook->new;
+}
+
+#else /** Version < v5.11 */
```
linD026, line 117: You used two different comment styles, `/** comment */` and `/* comment */`.
Please make sure they are consistent.

linD026 (changes requested):

lkmpg.tex
nickchen120235: I don't think I can rework this diagram in latex because it lacks features I need, like returning to functions other than caller (this is the most important one), annotation. (I'm not the creator of this sequence diagram so obtaining the "original" file is not possible either.)

I think the resolution of that jpeg is good enough for even printing, so I think I may left it as is.

<details><summary>The latex code I've written and the result</summary>

```tex
\begin{sequencediagram}
  \newthread{do_syscall_64}{do\_syscall\_64}
  \newinst[1.5]{sys_execve}{sys\_execve}
  \newinst[1.5]{ftrace}{[ftrace]}
  \newinst[1]{fh_ftrace_thunk}{fh\_ftrace\_thunk}
  \newinst[1]{fh_sys_execve}{fh\_sys\_execve}

      \postlevel \postlevel \postlevel
      \begin{call}{do_syscall_64}{\shortstack{
        \cpp|regs-ax=|\\
        \cpp|sys_call_table[nr]|\\
        \cpp|(regs->di,regs->si|\\
        \cpp|regs->dx,regs->r10|\\
        \cpp|regs->r8,regs->r9)|
      }}{sys_execve}{}
        \begin{call}{sys_execve}{call \cpp|__fentry__|}{ftrace}{}
          \begin{call}{ftrace}{}{fh_ftrace_thunk}{}
            \postlevel
          \end{call}
        \end{call}
        \begin{call}{sys_execve}{hooking}{fh_sys_execve}{\cpp|real_sys_execve()|}
          \postlevel
        \end{call}
        \postlevel
        \begin{call}{sys_execve}{call \cpp|__fentry__|}{ftrace}{}
          \begin{call}{ftrace}{}{fh_ftrace_thunk}{}
            \postlevel
          \end{call}
        \end{call}
        \begin{call}{sys_execve}{}{fh_sys_execve}{}
        \end{call}
      \end{call}
\end{sequencediagram}
```

image

</details>

linD026 (changes requested):

jserv (changes requested):
Rebase the latest 'master` branch.

PR #182, opened 2022-12-12
Add error message when sha256 algorithm is not supported

lyctw:
I just played with it using Linux 5.2.226 but got nothing prints on my screen, and realized I have to enable the config `CONFIG_CRYPTO_SHA256`. 

```
It failed silently when crypto_alloc_tfm() failed, so add an error message to inform the developer to enable sha256 algorithm support.

Signed-off-by: Yu Chien Peter Lin <peterlin.tw@pm.me>
```


jserv:
Thank @lyctw for contributing!

PR #183, opened 2022-12-25
example/syscall: Fix typo and unmatch types

linD026:


jserv:
Is there any automated approach to check unmatched types?

linD026:
I didn't use the Sparse (or other automated approaches), I just manually check [it](https://elixir.bootlin.com/linux/v6.1/source/fs/open.c#L1330).
And, I was surprised that Sparse didn't catch it.

PR #186, opened 2023-02-21
Discourage kernel.h usage and other header cleanups

ashevche:
We do not want to encourage people to use garbage which is kernel.h. The rule of thumb is to include the headers we are the direct users of. Clean up a bit the examples and text  of the book.

**UPDATE**
Besides that the examples have rather random headers. This has to be revisited and updated. In this PR it's a subset of the fixes of the mess.

### DIFF
--- a/examples/example_tasklet.c
+++ b/examples/example_tasklet.c
@@ -3,8 +3,8 @@
  */
 #include <linux/delay.h>
 #include <linux/interrupt.h>
-#include <linux/kernel.h>
 #include <linux/module.h>
+#include <linux/printk.h>
 
 /* Macro DECLARE_TASKLET_OLD exists for compatibility.
  * See https://lwn.net/Articles/830964/
--- a/examples/bottomhalf.c
+++ b/examples/bottomhalf.c
@@ -11,8 +11,8 @@
 #include <linux/delay.h>
 #include <linux/gpio.h>
 #include <linux/interrupt.h>
-#include <linux/kernel.h>
 #include <linux/module.h>
+#include <linux/printk.h>
 
 /* Macro DECLARE_TASKLET_OLD exists for compatibiity.
  * See https://lwn.net/Articles/830964/
--- a/examples/chardev.c
+++ b/examples/chardev.c
@@ -3,15 +3,19 @@
  * you have read from the dev file
  */
 
+#include <linux/atomic.h>
 #include <linux/cdev.h>
 #include <linux/delay.h>
 #include <linux/device.h>
 #include <linux/fs.h>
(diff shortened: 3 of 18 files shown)


jserv (changes requested):
Please run `clang-format -i examples/*.c` to enforce consistent indention.

linD026 (approved):
Looks good to me.

For the clang-format:
```
diff --git a/examples/hello-5.c b/examples/hello-5.c
index e2cb42c..3447855 100644
--- a/examples/hello-5.c
+++ b/examples/hello-5.c
@@ -2,7 +2,7 @@
  * hello-5.c - Demonstrates command line argument passing to a module.
  */
 #include <linux/init.h>
-#include <linux/kernel.h>      /* for ARRAY_SIZE() */
+#include <linux/kernel.h> /* for ARRAY_SIZE() */
 #include <linux/module.h>
 #include <linux/moduleparam.h>
 #include <linux/printk.h>
diff --git a/examples/intrpt.c b/examples/intrpt.c
index c35e02e..beeddd6 100644
--- a/examples/intrpt.c
+++ b/examples/intrpt.c
@@ -10,7 +10,7 @@
 
 #include <linux/gpio.h>
 #include <linux/interrupt.h>
-#include <linux/kernel.h>      /* for ARRAY_SIZE() */
+#include <linux/kernel.h> /* for ARRAY_SIZE() */
 #include <linux/module.h>
 #include <linux/printk.h>
```

ashevche:
@linD026 Thank you for review, I will update later. Maybe I will have time to check more examples.

jserv:
> @linD026 Thank you for review, I will update later. Maybe I will have time to check more examples.

Please resolve the compilation errors reported by GitHub Actions.

examples/example_tasklet.c
```diff
@@ -2,7 +2,6 @@
  * example_tasklet.c
  */
 #include <linux/delay.h>
-#include <linux/interrupt.h>
```
linD026, line 5: Do not remove this header since the example still uses the functions/macros, e.g., `tasklet_kill()`, in this header.

linD026 (changes requested):

linD026:
For the others, it looks good to me.

examples/example_tasklet.c
linD026: And, I think this is why GitHub Actions is yelling.

ashevche: Indeed. That's why as I said the examples needs to be revisited :-)
Thank you for your reviews!

jserv:
Thank @ashevche for contributing!

PR #213, opened 2023-07-23
procfs: Remove unnecessary cleanup on proc_create failure

keytouch:
I found it unneccesary to call `proc_remove` or `remove_proc_entry` when `proc_create` returns NULL.
I searched through kernel and found existing code never call remove on failed creates.
I read the kernel source of drivers/input/input.c

### DIFF
--- a/examples/procfs2.c
+++ b/examples/procfs2.c
@@ -77,7 +77,6 @@ static int __init procfs2_init(void)
 {
     our_proc_file = proc_create(PROCFS_NAME, 0644, NULL, &proc_file_fops);
     if (NULL == our_proc_file) {
-        proc_remove(our_proc_file);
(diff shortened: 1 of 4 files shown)


linD026 (approved):
Hi,
After I roughly look into the `proc_create` and `proc_remove`, I think `proc_remove` is unnecessary. (Even though, I found a use case in [here](https://elixir.bootlin.com/linux/v6.5-rc2/source/drivers/s390/block/dasd_proc.c#L336))

Thanks for the PR.

keytouch:
@linD026 Yes, this use case cleans up proc entries successfully created before the failed ones.
And it seems rather smart this way.

linD026:
Ah, I see.
My bad. Sorry for annoying you about this.

PR #215, opened 2023-08-04
Refine LaTeX package prerequisite

Urbaner3:
In line 30, it is sufficient to install xetex in order to save time, and to avoid compiling errors due to multi-version tex installation.

texlive-full includes all kinds of tex, saying luatex, xetex, pdflatex, and the installation takes time. 

Better to install texlive-xetex instead, after that `make` command works which creates the pdf file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ For Ubuntu Linux, macOS, and other Unix-like systems, run the following command(
 
 ```bash
 # Debian / Ubuntu
-$ sudo apt install make texlive-full
+$ sudo apt install make texlive-xetex
 
 # Arch / Manjaro
 $ sudo pacman -S make texlive-most texlive-bin


jserv:
@Hsins, can you help review?

Hsins:
Hi @Urbaner3,

You're right, one of the pains of existing $\LaTeX{}$ distributions is that they're often too big. The `texlive-full` package is a metapackage the pulls all the components and packages of TeX Live and almost occupies 5.x GB of disk space on installation.

> **Note**
> If I remember correctly, the lkmpg project is compiled with `pdflatex` instead of `xetex` so that the `texlive-xetex` could be replace with just `texlive`.

---

However, environments over people are complex and we leave the `texlive-full` to make sure that they have all the packages installed.

I would suggest not modifying the current instructions. It's better to add a description about "minimal installation" for those who need it. There're so many ways to approach:

- Install `TinyTeX` and install required packages with the TeX Live package manager `tlmgr`.  
(Check [TinyTeX](https://yihui.org/tinytex/) for more information)
- Install the package named `texlive-latex-base` and install required packages with the TeX Live package manager `tlmgr`.  
(Check [Basic Usage of `tlmgr`, the TeX Live Manager](http://tug.ctan.org/info/tlmgrbasics/doc/tlmgr.pdf) for more information)
- Install the package named `texlive`. It may contain the required packages needed for lkmpg.  
(Check [Difference between TeX Live Packages in Linux](https://tex.stackexchange.com/questions/245982/differences-between-texlive-packages-in-linux) for more information)
- Create a minimal docker image for usage.

Besides, the packages name `texlive`, `texlive-base`...etc are named and packaged by different package managers. For example, the Arch-packaged TeX Live named it differently with `texlive-basic`, `texlive-latex`, `texlive-luatex` ([REF](https://wiki.archlinux.org/title/TeX_Live)).

That would be weird if you just replaced `texlive-full` with `texlive` in the instructions for Debian/Ubuntu but still using `texlive-most` for Arch Linux and `mactex` for macOS.

Urbaner3:
After my study on 4 files, I can say that xetex is sufficient, since it depends texlive-base, which provides pdflatex. And since we use `latexmk` which specifies certain format of tex file, in lkmpg, the case is pdflatex, so the choises between tex running engines make no difference.
My option is similar to your third option, which is `install the package named texlive`. The conclusion is `xetek` is time saving choise.


Files:
man_latexmk.txt : `man latexmk`
tex-xetex.txt : `apt info texlive-xetek`
```
Depends: tex-common (>= 6.13), teckit, texlive-base (>= 2019.20200218)
```
tex-base.txt : `apt info texlive-base`
lshort.pdf : guidebook to ctan.org
https://ctan.org/tex-archive/info/lshort/english/
translation in Chinese, lshout_zh_cn.pdf

Hsins:
Have you ever checked the packages needed? There're the packages used:

```
cite, inputenc, fancyhdr, fncychap, fontenc, graphicx, 
hyperref, ifthen, minted, pdfpages, tikz, verbatim, xcolor, xparse
```

Note that the package `tikz` is not contained in `texlive-base` and `texlive-xetex`. Without using the `xelatex` engine for compile the document, why do we not just install the packages `texlive-pictures` directly? And how about the minimal installation guide for other platforms?

jserv:
I am inclined to close this pull request since it is difficult to determine how we can benefit from the proposed change.

PR #219, opened 2023-08-14
Update dependencies for Arch Linux

eeeXun:
From the [disscussion](https://bbs.archlinux.org/viewtopic.php?pid=2105974#p2105974), the `latexmk` has been moved to the `texlive-binextra` package.


jserv:
Thank @eeeXun for contributing!

PR #226, opened 2023-09-05
Correct name/location for ioctl-number.rst

mechanicalamit:
ioctl-number.rst in Documentation has been renamed and moved since v5.10

See commits
- 049500715e7
- 08536105d93


jserv:

PR #227, opened 2023-09-17
Chardev2 patches

mechanicalamit:


jserv (changes requested):
Refine the subject to address the motivations.

examples/chardev2.c
```diff
@@ -133,7 +134,7 @@ device_ioctl(struct file *file, /* ditto */
 
     /* Switch according to the ioctl called */
     switch (ioctl_num) {
-    case IOCTL_SET_MSG: {
```
linD026, line 136: No.
We must add the block scope to declare the variable in the specific case (statement). Otherwise, we might read the uninitialized memory.
It's not just the coding style issue, it is about the secure coding in C language.
See more information here: [EXP33-C. Do not read uninitialized memory](https://wiki.sei.cmu.edu/confluence/display/c/EXP33-C.+Do+not+read+uninitialized+memory)

linD026 (changes requested):

examples/chardev2.c
mechanicalamit: Ah, totally missed that.

mechanicalamit:
I will send a new pull request, closing this poorly done PR.

PR #222, opened 2023-09-01
Remove module * parameter from class_create()

mechanicalamit:
- From v6.4, class_create() does not have module *
  (See commit 11ba11 below)
- Added #ifdev for KERNEL_VERSION >= 6.4

From linux commit aaba11da9aa7d7d6b52a74d45b31cac118295a1
---------------------------------------------
driver core: class: remove module * from class_create()

The module pointer in class_create() never actually did anything,
and it shouldn't have been requred to be set as a parameter even
if it did something.  So just remove it and fix up all callers of the
function in the kernel tree at the same time.

PR #228, opened 2023-09-17
Remove module * parameter from class_create() 

mechanicalamit:
From v6.4, class_create() does not have module *
Same as (#222), commit 0addb868dc55 to chardev.c

### DIFF
--- a/examples/chardev2.c
+++ b/examples/chardev2.c
@@ -12,6 +12,7 @@
 #include <linux/printk.h>
 #include <linux/types.h>
 #include <linux/uaccess.h> /* for get_user and put_user */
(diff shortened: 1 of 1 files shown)


linD026:
The failure is from the [Sparse](https://git.kernel.org/pub/scm/devel/sparse/sparse.git/commit/?id=0e1aae55e49cad7ea43848af5b58ff0f57e7af99).
I'm fixing it right now.

jserv (changes requested):
Rebase the latest `master` branch where @linD026 resolved CI breakage.

jserv:

PR #232, opened 2023-09-22
Move kallsyms_lookup_name inside HAVE_KPROBES

mechanicalamit:
Draft PR looking for feedback.

kallsyms_lookup_name is defined inside #ifdef HAVE_KPROBES. Return call with kallsyms_lookup_name is outside the ifdef. Move kallsyms_lookup_name call inside the ifdef where the definition is.

Add a return NULL; call in the end, even though control might not reach there.

### DIFF
--- a/examples/syscall.c
+++ b/examples/syscall.c
@@ -178,9 +178,11 @@ static unsigned long **acquire_sys_call_table(void)
         return NULL;
     kallsyms_lookup_name = (unsigned long (*)(const char *name))kp.addr;
     unregister_kprobe(&kp);
-#endif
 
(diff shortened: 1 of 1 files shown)


linD026:
No,
the reason why we don't move kallsyms_lookup_name into the #ifdef
HAVE_KPROBES statement is kallsyms_lookup_name only be unexported since v5.7.

See more information from this commit https://github.com/sysprog21/lkmpg/commit/cccc98ab2c914cc922d34266b88ff73231e06607

mechanicalamit:
Ah, ok. Thank you for the detailed information and feedback.

PR #234, opened 2023-09-26
Simplify code by removal of outer struct encapsulating struct completions

mechanicalamit:
Draft PR looking for review and feedback.

Two struct completion(s) are encapsulated within another 'struct machine'. Simplify the code by removing the outer struct and let the struct completion(s) be self-standing.

For demonstrating an example, I believe the simpler code is better and does just as well. I am not sure if the encapsulation is necessary. Simplifying makes it easier to read and learn from.

### DIFF
--- a/examples/completions.c
+++ b/examples/completions.c
@@ -9,34 +9,32 @@
 #include <linux/printk.h>
 #include <linux/version.h>
 
-static struct {
-    struct completion crank_comp;
-    struct completion flywheel_comp;
-} machine;
+static struct completion crank_comp;
+static struct completion flywheel_comp;
 
 static int machine_crank_thread(void *arg)
 {
     pr_info("Turn the crank\n");
 
-    complete_all(&machine.crank_comp);
+    complete_all(&crank_comp);
 #if LINUX_VERSION_CODE >= KERNEL_VERSION(5, 17, 0)
-    kthread_complete_and_exit(&machine.crank_comp, 0);
+    kthread_complete_and_exit(&crank_comp, 0);
 #else
-    complete_and_exit(&machine.crank_comp, 0);
+    complete_and_exit(&crank_comp, 0);
 #endif
 }
 
 static int machine_flywheel_spinup_thread(void *arg)
 {
-    wait_for_completion(&machine.crank_comp);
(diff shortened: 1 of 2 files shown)


examples/completions.c
```diff
@@ -9,34 +9,32 @@
 #include <linux/printk.h>
 #include <linux/version.h>
 
-static struct {
-    struct completion crank_comp;
-    struct completion flywheel_comp;
-} machine;
```
linD026, line 15: You should also modify the passage in the Completions section:
> The machine structure stores the completion states for the two threads. At the exit point of each thread the respective completion state is updated, and wait_for_completion is used by the flywheel thread to ensure that it does not begin prematurely.

linD026 (changes requested):

examples/completions.c
mechanicalamit: Updated tex file to match code
-- Explained parts of a completions (init, wait_for_completion & complete)
-- Clarified the crank and flywheel threads better in the text

lkmpg.tex
```diff
@@ -1645,16 +1645,18 @@ \subsection{Completions}
 Sometimes one thing should happen before another within a module having multiple threads.
 Rather than using \sh|/bin/sleep| commands, the kernel has another way to do this which allows timeouts or interrupts to also happen.
 
-In the following example two threads are started, but one needs to start before another.
+Completions as code synchronization mechanism have three main parts, initialization of struct completion synchronization object, the waiting or barrier part through \cpp|wait_for_completion()|, and the signalling side through a call to \cpp|complete()|.
 
-\samplec{examples/completions.c}
-
-The \cpp|machine| structure stores the completion states for the two threads.
+In the following example two threads are started, but one thread(crank) needs to start before another thread(flywheel).
```
jserv, line 1650: Rephrase as following:
> In the subsequent example, two threads are initiated: crank and flywheel.
> It is imperative that the crank thread starts before the flywheel thread.
> A completion state is established for each of these threads, with a distinct completion defined for both the crank and flywheel threads.

mechanicalamit: Updated tex as requested.

jserv:

PR #235, opened 2023-09-28
Add newline to last pr_info to force dmesg to flush

mechanicalamit:
dmesg only flushes when it encounter a newline. Without a newline, the line is held in memory pending another printk. In this particular example (example_atomic.c), the last pr_info in atomic_bitwise() prints when another printk happens (either by another module, or __exit for this module.

This can be confusing to new learner. This patch adds a newline to the last pr_info forcing dmesg to print to the screen when the module is loaded.

--------------- Alternate solution ---------------
Alternate solutions could be adding a newline to BYTE_TO_BINARY_PATTERN
#define BYTE_TO_BINARY_PATTERN "%c%c%c%c%c%c%c%c\n"

... or adding a newline to each pr_info() call.

### DIFF
--- a/examples/example_atomic.c
+++ b/examples/example_atomic.c
@@ -50,7 +50,7 @@ static void atomic_bitwise(void)
     pr_info("Bits 4: " BYTE_TO_BINARY_PATTERN, BYTE_TO_BINARY(word));
 
     word = 255;
-    pr_info("Bits 5: " BYTE_TO_BINARY_PATTERN, BYTE_TO_BINARY(word));
+    pr_info("Bits 5: " BYTE_TO_BINARY_PATTERN "\n", BYTE_TO_BINARY(word));
 }
 
 static int __init example_atomic_init(void)


jserv:
I am not sure if adding a newline to each pr_info() call is better or not. I defer to @linD026 .

mechanicalamit:
On running `insmod example_atomic.ko`, the dmesg output is:
`[81692.796873] example_atomic started
[81692.796880] chris: 50, debbie: 52
[81692.796882] Bits 0: 00000000
[81692.796883] Bits 1: 00101000
[81692.796884] Bits 2: 00001000
[81692.796885] Bits 3: 00000000
[81692.796886] Bits 4: 00001000
`

Note the missing Bits 5 line in the dmesg output, even though that Bits 5 printk() has been called.

Then further on running `rmmod example_atomic.ko`, the Bits 5 line and example_atomic exit line will show up in the dmesg output.

To be honest, its not actually a code problem as you said above. The code is correct from a module point of view. At some point some other printk will occur and flush out the Bits 5 line.

The problem is when a learner (like me) tries the code, runs the insmod and wonders why the Bits 5 line did not output in dmesg. That is exactly what happened to me and frankly made me learn a few more things.

linD026 (approved):
Thanks for the explanation.
Now, I know what is this PR targeting.
It makes sense to me.

mechanicalamit:
Some amateur research (which might be incorrect).

In [printk.c](https://elixir.bootlin.com/linux/v6.5/source/kernel/printk/printk.c#L2214) prb_final_commit() or prb_commit() is called depending on the presence of a newline.

prb_*_commit functions are defined in[printk_ringbuffer.c](https://elixir.bootlin.com/linux/v6.5/source/kernel/printk/printk_ringbuffer.c#L1617)

From comments above the prb_*_commit functions, here are the relevant quotes.

- prb_commit()
> * Note that the data is not yet available to readers until it is finalized.
>  * Finalizing happens automatically when space for the next record is
>  * reserved
 
- prb_final_commit()
> * By finalizing, the data is made immediately available to readers.

So it might seem, that it is a printk behavior, and not a dmesg behavior as my commit message says.

printk also seems to remove newlines, and then adds its own.
[[ref 1](https://elixir.bootlin.com/linux/v6.5/source/kernel/printk/printk.c#L2130)][ref 2](https://elixir.bootlin.com/linux/v6.5/source/kernel/printk/printk.c#L1372)

mechanicalamit:
@jserv, the above was an attempt to understand and explain why this behavior was occurring. I don't think its an issue.

PR #236, opened 2023-10-06
Change word in tasklet description

mechanicalamit:
The word 'defator' is unable to be found in a lexicon. Change the word 'defator' with 'drawbacks' which should imply the negative meaning intended in the sentence.


jserv:

PR #237, opened 2023-10-07
Revise the depiction of return type of proc_create().

heartofrain:


jserv:
Thank @heartofrain for contributing!

PR #238, opened 2023-10-07
Fix grammar

visitorckw:
The verb 'suggest' is used with the gerund form.


jserv:
Thank @visitorckw for contributing!

PR #239, opened 2023-10-07
Replace schedule_work with queue_work

mechanicalamit:
schedule_work adds work to global workqueue. In this example, we create a local workqueue. Use the local workqueue by calling queue_work(), instead of putting work on the global workqueue.


linD026 (approved):

jserv:

PR #244, opened 2023-12-22
Fix compilation due to syscall module name conflict

keytouch:
Rename sys_call_table to _sys_call_table to fix compile error
sys_call_table is already declared in arch/x86/include/asm/syscall.h but of cource not exported by the kernel.

    /usr/src/linux-headers-6.1.0-16-common/arch/x86/include/asm/syscall.h:21:29:
    note: previous declaration of 'sys_call_table' with type 'long int (*
    const[])(const struct pt_regs *)'
       21 | extern const sys_call_ptr_t sys_call_table[];
And rename syscall module name to syscall_steal because In Debian, the name syscall conflicts with the patch debian/patches/features/x86/x86-make-x32-syscall-support-conditional.patch


examples/syscall_steal.c
```diff
@@ -1,5 +1,5 @@
 /*
- * syscall.c
+ * syscall_steal.c
```
linD026, line 2: > And rename syscall module name to syscall_steal because In Debian, the name syscall conflicts with the patch debian/patches/features/x86/x86-make-x32-syscall-support-conditional.patch

Could you please explain more details about the conflict?

examples/syscall_steal.c
```diff
@@ -61,7 +61,7 @@ module_param(sym, ulong, 0644);
 
 #endif /* Version < v5.7 */
 
-static unsigned long **sys_call_table;
+static unsigned long **_sys_call_table;
```
linD026, line 64: `_sys_call_table` is not a good name, if we do want to avoid the compiler warning we should have a better name not just add `_` or `__` prefix.

linD026 (changes requested):

examples/syscall_steal.c
keytouch: because of the [patch](https://lore.kernel.org/lkml/1415245982.3398.53.camel@decadent.org.uk/T/#u) in debian, there is a kernel parameter inside `/sys/module/syscall/`
I think the param prefix makes the name there and keeps another module with the same name from being loaded.
```
~/lkmpg/examples$ ll -R /sys/module/syscall/
/sys/module/syscall/:
total 0
drwxr-xr-x 2 root root    0 Dec 22 12:41 parameters
--w------- 1 root root 4096 Dec 21 14:39 uevent

/sys/module/syscall/parameters:
total 0
-r--r--r-- 1 root root 4096 Dec 22 12:41 x32
```
When I `insmod syscall.ko uid=xx`
```
~/lkmpg/examples$ sudo insmod syscall.ko uid=1000
insmod: ERROR: could not insert module syscall.ko: Invalid parameters
```
dmesg:
```
~/lkmpg/examples$ sudo dmesg | tail -n 1
[79500.176148] syscall: module is already loaded
```

examples/syscall_steal.c
keytouch: Shall we call it `stealed_sys_call_table`?😂

keytouch: And it is actually an error rather than a warning. I am on gcc 12.2.0.

linD026: > And it is actually an error rather than a warning. I am on gcc 12.2.0.

Sorry for my incorrect words.

> Shall we call it sys_call_table_stealed?😂

It is better than the `_sys_call_table` actually. LOL

keytouch: I edited my comment and I think `sys_call_table_acquired` might be better.

keytouch: But anyway it IS stolen. Let's still make it sys_call_table_stolen.

examples/syscall_steal.c
linD026: I see.
Then, you should add this information (including the patch source and mailing list) to the commit message.

linD026 (approved):

jserv:
Thank @keytouch for contributing!

PR #246, opened 2024-01-03
Change occurrences of egrep to grep -E

mechanicalamit:
A lot of distributions are emitting warnings when using egrep and fgrep since these are deprecated.

Update occurrences of egrep to 'grep -E' keeping the same functionality and making the code more portable without emitting warnings.


mechanicalamit:
.ci checks ran in 4 seconds since I believe it detected the sources files had not changed?

jserv:

PR #250, opened 2024-04-09
Replace SA_xxx IRQ flags with the IRQF_xxx ones

0xff07:
The SA_xxx flags has been removed for years. Nowadays the kernel uses the IRQF_xxx flags to specify IRQ behaviors. Adjust the descriptions in the book accordingly.


jserv:
Thank @0xff07 for contributing!

PR #251, opened 2024-04-13
Fix typo

weihsinyeh:
Correct "`hello world''" to "``hello world''".


jserv:
Thank @weihsinyeh for contributing!

PR #255, opened 2024-04-19
Rework the threaded IRQ examples

0xff07:
Rework descriptions for threaded IRQ, making it a section of its own. Also, fix errors in the `bh_threaded.c` example.

### DIFF
--- a/lkmpg.tex
+++ b/lkmpg.tex
@@ -1673,10 +1673,36 @@ The example below modifies the previous example to also run an additional task w
 
 \samplec{examples/bottomhalf.c}
 
-It is expected that at some unspecified point in the future tasklets may eventually be deprecated, to be replaced by \textit{threaded irqs} instead.
+\subsection{Threaded IRQ}
+
+Threaded IRQ is a mechanism to handle both top-half and bottom-half of an IRQ at once.
+A threaded IRQ splits one handler into two: one for the top-half, the other for the bottom-half.
+Those two handlers are registered at once by \cpp|request_threaded_irq()|.
+
+The top-half handler runs in interrupt context.
+It's the equivalence of the handler passed to the \cpp|request_irq()|.
+The bottom-half handler on the other hand runs in its own thread.
+This thread is created on registration of a threaded IRQ. Its sole purpose is to run this bottom-half handler.
+This is where a threaded IRQ is ``threaded''.
+
+Whether the bottom-half handler will be invoked is determined by the return value of the top-half handler.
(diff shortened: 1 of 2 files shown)


lkmpg.tex
```diff
@@ -1673,11 +1673,39 @@ \subsection{Bottom Half}
...
 \samplec{examples/bh_threaded.c}
 
+A threaded IRQ is registered using \cpp|request_threaded_irq()|.
+This function only takes one additional parameter than the \cpp|request_irq()| -- the bottom-half handling function that runs in its own thread.
+In this example it is the \cpp|button_bottom_half()|.
+Usage of other parameters are the same as \cpp|request_irq()|.
+
+Presence of both handlers is not mandatory.
+If either of them is not needed, pass the \cpp|NULL| instead.
+A \cpp|NULL| top-half handler implicitly means doing nothing but waking up the bottom-half serving thread (for running the bottom-half handler); A \cpp|NULL| bottom-half handler would have the same effect as if \cpp|request_irq()| were used.
```
jserv, line 1704: Split into two lines, delimited by a semicolon (`;`).

lkmpg.tex
```diff
@@ -1701,7 +1701,8 @@ \subsection{Threaded IRQ}
 
 Presence of both handlers is not mandatory.
 If either of them is not needed, pass the \cpp|NULL| instead.
-A \cpp|NULL| top-half handler implicitly means doing nothing but waking up the bottom-half serving thread (for running the bottom-half handler); A \cpp|NULL| bottom-half handler would have the same effect as if \cpp|request_irq()| were used.
+A \cpp|NULL| top-half handler implicitly means doing nothing but waking up the bottom-half serving thread (for running the bottom-half handler);
```
jserv, line 1704: Proofread as following:

A \cpp|NULL| top-half handler implies that no action is taken except to wake up the bottom-half serving thread, which runs the bottom-half handler.
Similarly, a \cpp|NULL| bottom-half handler effectively acts as if \cpp|request_irq()| were not used at all.

0xff07: Should be "as if \cpp|request_irq()| were used." instead of "as if \cpp|request_irq()| were not used at all."? I was trying to say that using:

```c
request_threaded_irq(irq, handler, NULL, flags, name, dev);
```

is effectively the same as:

```c
request_threaded_irq(irq, handler, flags, name, dev);
```

jserv: > Should be "as if \cpp|request_irq()| were used." instead of "as if \cpp|request_irq()| were not used at all."?

Agree. Please go ahead.

jserv:
Thank @0xff07 for contributing!

PR #257, opened 2024-05-05
Fix two typos in Preliminaries section

cvvletter:
Thank you for the helpful book! I found two typos in Sec. 5, which I hope to fix like this.


jserv:
Thank @cvvletter for contributing!

cvvletter:
That was quick! Happy to help.

Issue #52, opened 2021-08-09
Avoid using `try_module_get(THIS_MODULE)`


jserv:
There are various deprecated functions/macros used in LKMPG. For example, `try_module_get(THIS_MODULE)` is regarded as unsafe.

See https://stackoverflow.com/questions/1741415/linux-kernel-modules-when-to-use-try-module-get-module-put

ashevche:
I believe the book should state for which kernel versions it's applicable.

PR #258, opened 2024-05-08
vinput: Fix multiple issue

visitorckw:
Fix multiple issues for vinput module:

1. Removes redundant memset calls that were unnecessary since memory allocated with kzalloc is already zeroed due to the __GFP_ZERO gfp flag.
2. Fixes a NULL pointer dereference caused by failed kzalloc allocation. It adds a check for the return value of kzalloc and handles failed allocations by printing an error message and returning ERR_PTR(-ENOMEM).
3. Corrects the handling of failures in raw_copy_to_user() in vinput_read(). Previously, the function would modify '*offset' incorrectly on raw_copy_to_user() failure. Fix this behavior by changing count = -EFAULT to return -EFAULT instead.

### DIFF
--- a/examples/vinput.c
+++ b/examples/vinput.c
@@ -106,7 +106,7 @@ static ssize_t vinput_read(struct file *file, char __user *buffer, size_t count,
         count = len - *offset;
 
     if (raw_copy_to_user(buffer, buff + *offset, count))
-        count = -EFAULT;
+        return -EFAULT;
 
     *offset += count;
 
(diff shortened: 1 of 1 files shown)


jserv:
Can you check #52 ?

linD026:
Looks good to me.

visitorckw:
> Can you check #52 ?

I can spend some time understanding try_module_get and put_module_get and try to resolve this issue.

jserv:
Thank @visitorckw for contributing!

PR #260, opened 2024-06-20
Add system call "stealing" sample using kprobe handler

haodongnj:
The syscall-steal example does't work on my x84 Laptops with kernel 6.1 on Debian and 5.15 on Ubuntu.
As mentioned in  this [answer](https://stackoverflow.com/questions/78599971/hooking-syscall-by-modifying-sys-call-table-does-not-work/78607015#78607015) on stack overflow,  syscall table is no longer used to invoke system call on x86 arch after this [commit](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=1e3ad78334a69b36e107232e337f9d693dcc9df2), which is backed-ported to many LTS kernel versions. So maybe an extra example without using syscall table can make it easy to understand the situation for beginners like me .

### DIFF
--- a/examples/syscall-steal.c
+++ b/examples/syscall-steal.c
@@ -43,6 +43,14 @@
 
 #if defined(CONFIG_KPROBES)
 #define HAVE_KPROBES 1
+#if defined(CONFIG_X86_64)
+/* If you have tried to use the syscall table to intercept syscalls and it 
+ * doesn't work, you can try to use Kprobes to intercept syscalls.
+ * Set USE_KPROBES_PRE_HANDLER_BEFORE_SYSCALL to 1 to register a pre-handler
+ * before the syscall.
+ */
+#define USE_KPROBES_PRE_HANDLER_BEFORE_SYSCALL 0
+#endif
 #include <linux/kprobes.h>
 #else
 #define HAVE_PARAM 1
@@ -58,11 +66,36 @@ module_param(sym, ulong, 0644);
 
 #endif /* Version < v5.7 */
 
-static unsigned long **sys_call_table_stolen;
-
 /* UID we want to spy on - will be filled from the command line. */
 static uid_t uid = -1;
 module_param(uid, int, 0644);
+
+#if USE_KPROBES_PRE_HANDLER_BEFORE_SYSCALL
+
+/* syscall_sym is the symbol name of the syscall to spy on. The default is
+ * "__x64_sys_openat", which can be changed by the module parameter. You can 
+ * look up the symbol name of a syscall in /proc/kallsyms.
+ */
+static char *syscall_sym = "__x64_sys_openat";
+module_param(syscall_sym, charp, 0644);
+
+static int sys_call_kprobe_pre_handler(struct kprobe *p, struct pt_regs *regs)
+{
+    if (__kuid_val(current_uid()) != uid) {
+        return 0;
+    }
+
+    pr_info("%s called by %d\n", syscall_sym, uid);
+    return 0;
+}
+
+static struct kprobe syscall_kprobe = {
+    .symbol_name = "__x64_sys_openat",
+    .pre_handler = sys_call_kprobe_pre_handler,
+};
+#else
+
+static unsigned long **sys_call_table_stolen;
 
 /* A pointer to the original system call. The reason we keep this, rather
  * than call the original function (sys_openat), is because somebody else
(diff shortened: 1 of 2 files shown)


jserv (changes requested):
Merge the proposed changes of file `syscall-steal2.c` into `syscall-steal.c` so that the content can appear in the book.

jserv:
Reported by CI:
> examples/syscall-steal2.c:35:15: warning: symbol 'syscall_kprobe' was not declared. Should it be static?

examples/syscall-steal.c
```diff
@@ -43,6 +43,15 @@
 
 #if defined(CONFIG_KPROBES)
 #define HAVE_KPROBES 1
+#if defined(CONFIG_X86_64)
+/* For x86 architecture, syscall table can no longer be used to invoke a syscall
```
jserv, line 47: Rewrite the comment as following:
```c
 /* For x86 architecture, the syscall table cannot be used to invoke a syscall
 * after commit 1e3ad78.
 * Use a hook on the syscall entry instead to intercept the syscall.
 * Alternatively, set USE_KPROBES_PRE_HANDLER_BEFORE_SYSCALL to 0 to use the old method.
 */
```

examples/syscall-steal.c
```diff
@@ -43,6 +43,14 @@
 
 #if defined(CONFIG_KPROBES)
 #define HAVE_KPROBES 1
+#if defined(CONFIG_X86_64)
+ /* For x86 architecture, the syscall table cannot be used to invoke a syscall
+ * after commit 1e3ad78.
+ * Use a hook on the syscall entry instead to intercept the syscall.
+ * Alternatively, set USE_KPROBES_PRE_HANDLER_BEFORE_SYSCALL to 0 to use the old method.
+ */
+#define USE_KPROBES_PRE_HANDLER_BEFORE_SYSCALL 1
```
jserv, line 52: You should set `USE_KPROBES_PRE_HANDLER_BEFORE_SYSCALL` to zero for the opposite condition.

examples/syscall-steal.c
linD026: Adding the description of the kernel version, i.e., `... after v6.9 (commit 1e3ad78) ...`, would be better.

examples/syscall-steal.c
```diff
@@ -43,6 +43,14 @@
 
 #if defined(CONFIG_KPROBES)
 #define HAVE_KPROBES 1
+#if defined(CONFIG_X86_64)
+ /* For x86 architecture, the syscall table cannot be used to invoke a syscall
+ * after commit 1e3ad78.
```
linD026, line 48: Adding the description of the kernel version, i.e., `... after v6.9 (commit 1e3ad78) ...`, would be better.

examples/syscall-steal.c
```diff
@@ -43,6 +43,14 @@
...
+ * Use a hook on the syscall entry instead to intercept the syscall.
+ * Alternatively, set USE_KPROBES_PRE_HANDLER_BEFORE_SYSCALL to 0 to use the old method.
+ */
+#define USE_KPROBES_PRE_HANDLER_BEFORE_SYSCALL 1
+#endif
```
linD026, line 53: Like:
```cpp
#ifdef ...
#else
#define USE_KPROBES_PRE_HANDLER_BEFORE_SYSCALL 0
#endif
```

examples/syscall-steal.c
```diff
@@ -43,6 +43,14 @@
...
 #include <linux/kprobes.h>
 #else
 #define HAVE_PARAM 1
```
linD026, line 56: Also, we need to add the other condition, the system is v5.9+, and x86 arch without kprobe support.
Maybe we can tell the user they are unable to run this example.

linD026 (changes requested):

examples/syscall-steal.c
haodongnj: This commit has been backported to long term stable kernels, like v5.15, v6.1, v6.6 and v6.8, which is added in comment for now.

examples/syscall-steal.c
haodongnj: Maybe defining USE_KPROBES_PRE_HANDLER_BEFORE_SYSCALL to 0 by default is enough because users may turn on this macro if they get stuck when hacking.

linD026: But the module will be broken if we have such a scenario.
Moreover, the user may be confused about this and start googling your question again.
I would suggest that you provide a more comprehensive explanation or relevant links (e.g., the stack overflow answer you mentioned).
Additionally, in the code, we can print out some warnings about this.

examples/syscall-steal.c
```diff
@@ -43,6 +43,19 @@
 
 #if defined(CONFIG_KPROBES)
 #define HAVE_KPROBES 1
+#if defined(CONFIG_X86_64)
+/* For x86 architecture, the syscall table cannot be used to invoke a syscall
+ * after commit 1e3ad78 since v6.9. This commit has been backported to long 
+ * term stable kernels, like v5.15, v6.1, v6.6 and v6.8. In this case, use a 
+ * hook on the syscall entry instead to intercept the syscall. 
+ * For more details, see https://stackoverflow.com/questions/78599971/hooking-syscall-by-modifying-sys-call-table-does-not-work/78607015#78607015.
```
jserv, line 51: The rationale for using Kprobe on the x86 architecture is informative. Could you move the description out of this file? Specifically, the text should appear in the LaTeX script so that readers can gain a deeper understanding before examining the kernel module source code.

Xxblimp:
Hey I'm new here and I want to make a friend who will guide me through coding

jserv (changes requested):
Read https://cbea.ms/git-commit/ carefully and squash the git commits.

jserv:
Thank @haodongnj for contributing!

PR #262, opened 2024-07-02
Modify the descritions of seq_file and flowchart

jouae:
7.4 The seq_file is missing the behavior that occurs when it directly returns NULL after the start step.

Also, in the flowchart, the first decision where `return is NULL` is `Yes` points to nothing, which is not defined in ISO 5807.

ISO 5807 section 10.3.1.2 states logic path of a decision, "Each exit from a symbol be accompanied by the appropriate condition values to show the logic path which it represents, so that the conditions and the associated references are identified."

The behavior of pointing to nothing does not have associated references and does not indicate the relation of this logic path.

Refs:
	1. The Liunx Kernel Module Programming Guide 7.4 Manage /proc file with seq_file
	2. ISO-5807-1985

### DIFF
--- a/lkmpg.tex
+++ b/lkmpg.tex
@@ -1234,7 +1234,7 @@ It is based on sequence, which is composed of 3 functions: \cpp|start()|, \cpp|n
 The \cpp|seq_file| API starts a sequence when a user read the \verb|/proc| file.
 
 A sequence begins with the call of the function \cpp|start()|.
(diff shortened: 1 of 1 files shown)


jserv:
Thank @jouae for contributing!

jouae:
I find a typo duplicated word "otherwise",

let me fix it when I get off work today.

PR #263, opened 2024-07-02
docs: Update example in chapter 8 to use sudo

pao0626:
In chapter 8, the example code for setting permissions is as follows:
```c
static struct kobj_attribute myvariable_attribute =
    __ATTR(myvariable, 0660, myvariable_show, (void *)myvariable_store);
```
We have two options:
1. Instruct users to access the file with `sudo`.
2. Relax the permission setting to 0666.

This commit adopts the sudo method to maintain security constraints.

### DIFF
--- a/lkmpg.tex
+++ b/lkmpg.tex
@@ -1347,14 +1347,14 @@ sudo lsmod | grep hello_sysfs
 What is the current value of \cpp|myvariable| ?
 
 \begin{codebash}
-cat /sys/kernel/mymodule/myvariable
+sudo cat /sys/kernel/mymodule/myvariable
 \end{codebash}
 
 Set the value of \cpp|myvariable| and check that it changed.
 
 \begin{codebash}
-echo "32" > /sys/kernel/mymodule/myvariable
-cat /sys/kernel/mymodule/myvariable
+echo "32" | sudo tee /sys/kernel/mymodule/myvariable
+sudo cat /sys/kernel/mymodule/myvariable
 \end{codebash}
 
 Finally, remove the test module:


lkmpg.tex
```diff
@@ -1347,14 +1347,14 @@ \section{sysfs: Interacting with your module}
...
-cat /sys/kernel/mymodule/myvariable
+sudo cat /sys/kernel/mymodule/myvariable
 \end{codebash}
 
 Set the value of \cpp|myvariable| and check that it changed.
 
 \begin{codebash}
-echo "32" > /sys/kernel/mymodule/myvariable
-cat /sys/kernel/mymodule/myvariable
+sudo sh -c 'echo "32" > /sys/kernel/mymodule/myvariable'
```
jserv, line 1356: Use the combination of `sudo` and `tee` commands. See https://askubuntu.com/questions/408340/is-there-any-significance-to-using-tee

lkmpg.tex
```diff
@@ -1347,14 +1347,14 @@ \section{sysfs: Interacting with your module}
...
+sudo cat /sys/kernel/mymodule/myvariable
```
jserv, line 1357: See if `sudo cat` is necessary or not.

lkmpg.tex
pao0626: Thank you for your suggestion. I have updated the command to use the combination of sudo and tee to perform the write operation. Here is the revised command:
```c
echo "32" | sudo tee /sys/kernel/mymodule/myvariable
```
This ensures that the data is correctly written to the file requiring elevated permissions.

lkmpg.tex
pao0626: To address your question about whether sudo cat is necessary: Since the file permissions for /sys/kernel/mymodule/myvariable are set to 0660, it means that only the owner and group members have read and write access, while others have no access. Therefore, if the current user is not the owner or part of the group that owns the file, sudo cat would indeed be necessary to read the file's contents.

However, if the permissions are changed to 0664, then cat would not require sudo since others would have read access. Similarly, as I mentioned initially, if the permissions are set to 0666, both read and write operations would not require sudo.

Please let me know if there are any additional changes or clarifications needed.

pao0626: I have also confirmed through my own testing that sudo cat is necessary with the current permissions.

jserv (changes requested):
Read https://cbea.ms/git-commit/ carefully and refine.

pao0626:
I actually wrote my initial commit message after carefully reviewing the content from the link you provided (https://cbea.ms/git-commit/), and also referred to https://www.freecodecamp.org/news/how-to-write-better-git-commit-messages/ for adding a <type> to the title. However, I couldn't find the best way to express it.

I have now rethought and submitted a new commit message. If there are still areas that can be improved, please let me know.

jserv (changes requested):
Don't specify the title as "Update lkmpg.tex" since we have one Tex document file. You have to summarize in more informative manner.

jserv:
Thank @pao0626 for contributing!

PR #264, opened 2024-07-03
Fix a typo

jouae:
Delete the duplicate word "otherwise"

Before and after:
otherwise, otherwise -> otherwise


jserv:
Thank @jouae for contributing!

PR #98, opened 2021-09-08
Adapt timer API usage

linD026:
Since v4.14[1], the timer API started to change.
The series of improvements ended up at v4.15[2].
We have to update the API.

Related article: https://lwn.net/Articles/735887/

[1] https://github.com/torvalds/linux/commit/686fef928bba6be13cabe639f154af7d72b63120
[2] https://github.com/torvalds/linux/commit/841b86f3289dbe858daeceec36423d4ea286fac2

PR #265, opened 2024-07-21
Correct return-object to return-oriented programming

classAndrew:
Change `return-object programming` to `return-oriented programming` in timer API section.

This was mentioned [here](https://github.com/sysprog21/lkmpg/pull/98#discussion_r704877998) but didn't make it into the final revision.


jserv:
Thank @classAndrew for contributing!

PR #266, opened 2024-08-11
Fix: blocking O_NONBLOCK process bug in sleep.c

Ayxan13:
There is a subtle bug that if the atomic flag changes between the time it was checked and the second time it was checked, sleep.c would potentially block a process that had specified `O_NONBLOCK`. This fixes the bug by using `atomic_cmpxchg` instead of `atomic_read`.


linD026 (approved):

jserv:
Thank @Ayxan13 for contributing!

PR #271, opened 2024-10-06
Remove the redundant code

NOVBobLee:
proc_remove does nothing when the argument is NULL[1]. We don't need proc_remove after proc_create failed.

[1] https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/fs/proc/generic.c#n789


jserv:

PR #272, opened 2024-10-06
Fix a wierd behavior of a function

NOVBobLee:
The procfile_write prints the content what user writes into. However, when the content size is greater than or equal to PROCFS_MAX_SIZE, procfile_write will print nothing, because the index for appending the tail NULL character will be modulo to 0.

So, I modify the code with 1. restrict the max acceptable content size to (PROCFS_MAX_SIZE - 1), left a space for NULL character 2. discard the modulo because we already restrict the max size to (PROCFS_MAX_SIZE - 1)

### DIFF
--- a/examples/procfs2.c
+++ b/examples/procfs2.c
@@ -48,13 +48,13 @@ static ssize_t procfile_write(struct file *file, const char __user *buff,
                               size_t len, loff_t *off)
 {
     procfs_buffer_size = len;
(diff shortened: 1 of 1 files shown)


jserv (changes requested):
Refine the git commit message (including the typo: "wierd") and address the [off-by-one error](https://en.wikipedia.org/wiki/Off-by-one_error).

linD026 (approved):

jserv:

PR #273, opened 2024-10-07
Correct the function type

NOVBobLee:
The function pointer type of myvariable_store is incompatible with the store member of struct kobj_attribute[1]. So, if we omit the type conversion (void *), we will meet a compilation error. This patch corrects the function declaration and removes the type conversion.

[1] https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/linux/kobject.h#n139

### DIFF
--- a/examples/hello-sysfs.c
+++ b/examples/hello-sysfs.c
@@ -20,7 +20,7 @@ static ssize_t myvariable_show(struct kobject *kobj,
 }
 
 static ssize_t myvariable_store(struct kobject *kobj,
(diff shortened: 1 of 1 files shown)


jserv (changes requested):
Improve the English writing in git commit messages, resolving grammar mistakes and making the access to structure members more readable.

NOVBobLee:
Just refine the git commit message. Please review it again, thx!

jserv:

PR #276, opened 2024-10-20
Fix incorrect failure handling in vinput_init

NOVBobLee:
After class_register failure, the wrong function class_unregister was used. The correct cleanup function should be unregister_chrdev, as register_chrdev was used before class_register.

### DIFF
--- a/examples/vinput.c
+++ b/examples/vinput.c
@@ -401,7 +401,7 @@ static int __init vinput_init(void)
 
     return 0;
 failed_class:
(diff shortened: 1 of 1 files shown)


jserv:
Thanks for spotting this issue. Can you point out more in current example code?

NOVBobLee:
Yes, I’ll go through the current example code and point out any additional issues I find.

linD026 (approved):

jserv:

PR #278, opened 2024-10-24
Fix potential reference count leak in vinput

NOVBobLee:
In the export_store function, when vinput_register_vdevice fails, the current error handling process calls device_unregister followed by vinput_destroy_vdevice. However, the device_unregister here triggers its release function, which already includes a call to vinput_destroy_vdevice. As a result, vinput_destroy_vdevice is called twice. Since vinput_destroy_vdevice contains module_put, this double call must be avoided to prevent potential reference count issues.


NOVBobLee:
Fixed typo.

linD026 (approved):

jserv:

PR #280, opened 2024-11-05
Fix missing function call in error handling path

NOVBobLee:
Previously, the code did not call kobject_put when sysfs_create_file
failed, leading to a potential memory leak. This commit adds the missing
kobject_put to ensure proper resource cleanup.


jserv:

PR #281, opened 2024-11-11
Replace make with $(MAKE) in Makefile

Integral-Tech:
According to [GNU Make Manual](https://www.gnu.org/software/make/manual/html_node/MAKE-Variable.html),

Recursive make commands should always use the variable `MAKE`, not the explicit command name `make`.


jserv:
Thank @Integral-Tech for contributing!

PR #284, opened 2024-11-12
Add missing close() and use EXIT_FAILURE macro in cat_nonblock.c

Integral-Tech:
In `cat_nonblock.c`, function call of `close()` is missing, which could lead to resource leak. Besides, using `EXIT_FAILURE` macro defined in stdlib.h provides better readability.


jserv:
Thank @Integral-Tech for contributing!

PR #289, opened 2024-12-11
Allow devicemodel built with Linux v6.11+

jeremy90307:
Since commit 0edb555 ("platform: Make platform_driver::remove() return void"), the original `platform_driver::remove()` function returned an integer value, usually an error code. This made driver developers think that errors were properly handled. However, the driver core does not stop the device removal process based on the error code, so the error code did not serve its intended purpose. Therefore, the return type was changed to `void` to avoid potential issues.

Make the devicemodel example run successfully on Linux v6.11+.

### DIFF
--- a/examples/devicemodel.c
+++ b/examples/devicemodel.c
@@ -4,6 +4,7 @@
 #include <linux/kernel.h>
 #include <linux/module.h>
 #include <linux/platform_device.h>
+#include <linux/version.h>
 
 struct devicemodel_data {
     char *greeting;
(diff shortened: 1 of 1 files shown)


jeremy90307:
Here are the errors I encountered when compiling devicemodel.c on Linux v6.12.1.

`/devicemodel.c:70:15: error: initialization of ‘void (*)(struct platform_device *)’ from incompatible pointer type ‘int (*)(struct platform_device *)’ [-Werror=incompatible-pointer-types]
   70 |     .remove = devicemodel_remove,
      |               ^~~~~~~~~~~~~~~~~~
/home/jeremypi/tmp/test_com/devicemodel.c:70:15: note: (near initialization for ‘devicemodel_driver.<anonymous>.remove’)`

jserv:
Thank @jeremy90307 for contributing!

PR #308, opened 2025-04-20
Improve clarity in file access example

EricccTaiwan:
Replaced "vt" with "virtual terminal" to improve readability, as the abbreviation appears for the first time in this section. Also rephrased to avoid nested parentheses.


jserv:
Thank @EricccTaiwan for contributing!

PR #311, opened 2025-04-24
Align struct file_operations with upstream v6.14

Jordymalone:
Chapter 6, "Character Device Drivers", section 6.1 still showed the v5.4 layout of 'struct file_operations'. Replace it with the version shipped in upstream Linux v6.14, keeping the example buildable and the API in step with modern kernels.

Upstream reference: https://elixir.bootlin.com/linux/v6.14/source/include/linux/fs.h#L2131

### DIFF
--- a/lkmpg.tex
+++ b/lkmpg.tex
@@ -908,46 +908,52 @@ Each field of the structure corresponds to the address of some function defined 
 
 For example, every character driver needs to define a function that reads from the device.
 The \cpp|file_operations| structure holds the address of the module's function that performs that operation.
-Here is what the definition looks like for kernel 5.4:
+Here is what the definition looks like for kernel 6.14:
 
 \begin{code}
 struct file_operations {
     struct module *owner;
+    fop_flags_t fop_flags;
     loff_t (*llseek) (struct file *, loff_t, int);
     ssize_t (*read) (struct file *, char __user *, size_t, loff_t *);
     ssize_t (*write) (struct file *, const char __user *, size_t, loff_t *);
     ssize_t (*read_iter) (struct kiocb *, struct iov_iter *);
     ssize_t (*write_iter) (struct kiocb *, struct iov_iter *);
-    int (*iopoll)(struct kiocb *kiocb, bool spin);
-    int (*iterate) (struct file *, struct dir_context *);
+    int (*iopoll)(struct kiocb *kiocb, struct io_comp_batch *,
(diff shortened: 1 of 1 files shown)


lkmpg.tex
```diff
@@ -908,46 +908,52 @@ \subsection{The file\_operations Structure}
 
 For example, every character driver needs to define a function that reads from the device.
 The \cpp|file_operations| structure holds the address of the module's function that performs that operation.
-Here is what the definition looks like for kernel 5.4:
+Here is what the definition looks like for kernel 6.14:
```
jserv, line 911: Don't do that. Linux v5.4 is a l[ong-term support kernel](https://www.kernel.org/category/releases.html). LKMPG maintains compatibility with LTS kernels, which means all content should be compatible with Linux v5.4 and newer versions, rather than focusing exclusively on recent Linux releases. Instead of modifying the struct definitions, please document any changes to member fields in the descriptions.

Jordymalone: Thanks for your clarification! That make sense - I’ll revert the `struct file_operations` example to the original Linux 5.4 definition (undoing my earlier changes) so it stays compatible with all LTS kernels.

To cover newer kernels without breaking the build, I’ll add a short note right after the code that lists the fields added, removed, or modified since 5.4.

> As of Linux kernel 6.12, several member fields have been added or removed, or had their prototypes changed:
> - **Added**: `fop_flags`, `splice_eof`, `uring_cmd`, `uring_cmd_iopoll`, `mmap_capabilities`
> - **Removed**: `iterate`, `sendpage`, `mmap_supported_flags` (!CONFIG_MMU)
> - **Modified**: 
>   - `iopoll` now takes `struct io_comp_batch *` and `unsigned int flags`
>   - `setlease` now uses `struct file_lease **` and `int` instead of `long`

Let me know if this approach fits the documentation style you prefer, or if you'd suggest any adjustments.

jserv: Go ahead on top of Linux v5.4 codebase.

jserv:
Any progress?

Jordymalone:
> Any progress?

Thanks for the ping and sorry for the long delay.

I've just pushed the updated version with the revised commit message as we discussed. The main text now remains on the v5.4 baseline, and a footnote has been added to summarize the API changes in modern kernels.

jserv:
Thank @Jordymalone for contributing!

PR #342, opened 2025-09-04
Fix itemize indentation

Jordymalone:
Remove extra spaces before `\item` under `\begin{itemize}` to normalize indentation. This change is formatting-only and does not alter any text content. Normalized style reduces unnecessary diff noise and improves consistency across the document.


jserv:
The indention should be restored.

Jordymalone:
> The indention should be restored.

Understood. But I noticed the file mixes different indentation styles for `\item`.
Should I fully restore everything as before?
Or keep `\begin{itemize}`/`\end{itemize}` left-aligned and normalize all `\item` to a consistent indent (e.g., two spaces)?

EricccTaiwan:
> > The indention should be restored.
> 
> Understood. But I noticed the file mixes different indentation styles for `\item`. Should I fully restore everything as before? Or keep `\begin{itemize}`/`\end{itemize}` left-aligned and normalize all `\item` to a consistent indent (e.g., two spaces)?

I think it would be nice to keep the formatting consistent.

jserv:
Or keep `\begin{itemize}`/`\end{itemize}` left-aligned and normalize all `\item` to a consistent indent (e.g., two spaces)?

The preferable style would be:
```tex
\begin{enumerate}
  \item The application loads the syscall number into a CPU register (e.g., \verb|EAX| on x86)                                                         
  \item A special instruction triggers the transition (\verb|int 0x80| or \verb|sysenter| on x86)
  \item The CPU switches to kernel mode and jumps to the syscall handler
  \item The kernel validates arguments, performs the operation, and returns the result
  \item Control returns to user mode with the result in a register
\end{enumerate}
```

Use two space indention.

jserv:
Thank @Jordymalone for contributing!

Issue #348, opened 2025-09-17
Rationale for removing .owner = THIS_MODULE in struct file_operations (Linux v6.4+)


visitorckw:
The conditional removal of the explicit .owner = THIS_MODULE initialization in recent module examples was introduced in commit 1c31bac ("Fix Linux v6.5 compatibility"):

```c
static struct file_operations fops = {
#if LINUX_VERSION_CODE < KERNEL_VERSION(6, 4, 0)
    .owner = THIS_MODULE,
#endif
    // ...
};
```

My Questions:

1. What is the specific Linux kernel commit that introduced this change, and what is the technical rationale? (Since struct file_operations::owner still exists in the header, I assume the core kernel now automatically handles module usage counting for file operations, making the explicit setting redundant?)

2. The change was introduced to fix "Linux v6.5 compatibility," but the version check uses < 6.4.0. Could you please clarify the exact version boundary, and whether the condition should potentially be adjusted (e.g., changing < to <= or the version numbers)?

visitorckw:
> > 1. What is the specific Linux kernel commit that introduced this change, and what is the technical rationale? (Since struct file_operations::owner still exists in the header, I assume the core kernel now automatically handles module usage counting for file operations, making the explicit setting redundant?)
> 
> I actually made a similar PR in [sysprog21/simrupt#12](https://github.com/sysprog21/simrupt/pull/12) , and I think the commit message there already explains your question pretty well!

I don't think that's the correct commit.

The Greg's patch you mentioned cleans up struct class, not struct file_operations. This still doesn't explain why it's safe for file_operations.

From a quick search, several in-tree file_operations definitions still explicitly set .owner = THIS_MODULE.

> 
> > 2. The change was introduced to fix "Linux v6.5 compatibility," but the version check uses < 6.4.0. Could you please clarify the exact version boundary, and whether the condition should potentially be adjusted (e.g., changing < to <= or the version numbers)?
> 
> No, because the change was first introduced in `v6.4-rc1`, so I believe keeping `< 6.4` is correct. That said, I agree the commit title is a bit misleading.

OK, I'll proceed with the understanding that the previous commit message description was simply incorrect.

Additionally, even if this change is technically correct, I still doubt the necessity of adding this conditional compilation. It seems setting .owner remains correct and safe in all kernel versions. Introducing a conditional compile block like this only serves to confuse readers, leading them to question if struct file_operations even has the .owner member after v6.4.

jserv:
[...]

PR #349, opened 2025-09-18
Fix incorrect conditional usage of .owner

visitorckw:
The examples code wrapped .owner = THIS_MODULE inside version checks, suggesting that it is no longer needed after v6.4. This is only true for driver types where the driver core sets .owner automatically (e.g. platform and i2c drivers).

For structures such as struct file_operations, .owner must still be set explicitly by the user. Without it, module reference counting will be broken, allowing a module to be unloaded while still in use, which can lead to kernel panics.

struct class is a special case: its .owner field was removed in upstream Linux commit 6e30a66433af ("driver core: class: remove struct module owner out of struct class"). Explicitly setting .owner for struct class is safe on older kernels but must be conditionally compiled out on newer kernels.

Remove the unnecessary version guards and unconditionally sets .owner = THIS_MODULE in the affected example code.

Closes: https://github.com/sysprog21/lkmpg/issues/348
    

---

## Summary by cubic
Remove version guards and always set .owner = THIS_MODULE in example drivers to fix module refcounting and avoid unload panics. This also makes the examples clearer.

- **Bug Fixes**
  - Always set .owner in struct file_operations and cdev; removed LINUX_VERSION_CODE checks to ensure correct refcounting.
  - Keep .owner in struct class unguarded; optional but correct and less confusing.

### DIFF
--- a/examples/dht11.c
+++ b/examples/dht11.c
@@ -148,12 +148,10 @@ static ssize_t device_read(struct file *filp, char __user *buffer,
 }
 
 static struct file_operations fops = {
-#if LINUX_VERSION_CODE < KERNEL_VERSION(6, 4, 0)
     .owner = THIS_MODULE,
-#endif
     .open = device_open,
     .release = device_release,
-    .read = device_read
+    .read = device_read,
 };
 
 /* Initialize the module - Register the character device */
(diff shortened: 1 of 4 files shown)


visitorckw:
Hmm. So it looks like the new kernel even removed the .owner field from struct class? I’ll double-check.

PR #359, opened 2026-02-10
Fix retrieve user value in IOCTL_VALSET_NUM case

applepwc:
The test program is:

``` c
    arg.val = 0xAA;
    ioctl(fd, IOCTL_VALSET_NUM, &arg);

    arg.val = 0;
    ioctl(fd, IOCTL_VALGET_NUM, &arg);
    printf("Read val: 0x%x\n", arg.val);
```

---
## Summary by cubic
Switch IOCTL_VALSET_NUM to _IO so the integer is passed via the ioctl arg, matching the driver handler and example calls. This fixes the set/get number flow and keeps the convention consistent across the codebase.

<sup>Written for commit 113cd574bb8b47df4fbc57e5a15eda3873068f6f. Summary will update on new commits.</sup>

### DIFF
--- a/examples/ioctl.c
+++ b/examples/ioctl.c
@@ -20,7 +20,7 @@ struct ioctl_arg {
 #define IOCTL_VALSET _IOW(IOC_MAGIC, 0, struct ioctl_arg)
 #define IOCTL_VALGET _IOR(IOC_MAGIC, 1, struct ioctl_arg)
 #define IOCTL_VALGET_NUM _IOR(IOC_MAGIC, 2, int)
-#define IOCTL_VALSET_NUM _IOW(IOC_MAGIC, 3, int)
+#define IOCTL_VALSET_NUM _IO(IOC_MAGIC, 3)
 
 #define IOCTL_VAL_MAXNR 3
 #define DRIVER_NAME "ioctltest"


EricccTaiwan (approved):

jserv (changes requested):
Remove "fix: " in the subject of git commit messages.
Refine git commit message, making it more informative. See https://chris.beams.io/git-commit
Don't close this pull request. Use git force-push instead.

examples/ioctl.c
```diff
@@ -75,7 +75,10 @@ static long test_ioctl_ioctl(struct file *filp, unsigned int cmd,
         break;
 
     case IOCTL_VALSET_NUM:
-        ioctl_num = arg;
+        int user_val;
```
linD026: To be compatible with the old version, which might still use the C89 standard.
Don't declare the variable inside the switch statement.
Or, you should declare the variable within the block scope.

Also, if you want to declare the variable inside the switch statement, please fix this:
```
ioctl.c:78:9: error: typename in expression
ioctl.c:78:13: error: Expected ; at end of statement
ioctl.c:78:13: error: got user_val
ioctl.c:78:9: error: undefined identifier 'int'
ioctl.c:79:13: error: undefined identifier 'user_val'
ioctl.c:81:21: error: undefined identifier 'user_val'
```

linD026 (changes requested):

applepwc:
Hello, I force-pushed a new commit. Please review it.

visitorckw:
> Hello, I force-pushed a new commit. Please review it.

Github notifies the relevant people automatically, so there's no need to leave a comment like this.

visitorckw:
Thanks for the patch.

While this addresses the pointer-to-value mismatch, should we consider whether `IOCTL_VALSET_NUM` is better suited for "pass-by-value" rather than "pass-by-reference"?

Currently, the macro is defined as:
```
#define IOCTL_VALSET_NUM _IOW(IOC_MAGIC, 3, int)
```

Using `_IOW `implies that arg is a user-space pointer. However, for a simple integer assignment, would it be more idiomatic to use pass-by-value by redefining the macro with `_IO`? This would allow us to keep the logic lightweight without the overhead of memory copying.

jserv (changes requested):
Ensure consistent code style with 'clang-format' in `examples` directory.

applepwc:
Yes, using _IO with pass-by-value is more idiomatic and avoids unnecessary memory copying.
But this would be an ABI change: existing user-space code calling ioctl(fd, IOCTL_VALSET_NUM, &value) would break.

visitorckw:
Wouldn't the proposed patch also change the existing ABI?

applepwc:
Hello, I'm uncertain whether to add more code as shown below or simply rewrite the original.

``` c
#define IOCTL_VALSET_NUM2 _IO(IOC_MAGIC, 4)
#define IOCTL_VAL_MAXNR 4
...
case IOCTL_VALSET_NUM2:
    ioctl_num = (int)arg; 
    break;
```

linD026:
Based on the `userspace_ioctl.c`:

```c
int ioctl_get_nth_byte(int file_desc)  
{  
    int i, c; 

    printf("get_nth_byte message:");
    i = 0;  
    do {  
        c = ioctl(file_desc, IOCTL_GET_NTH_BYTE, i++);
    /* ... */
```
The test program in the LKMPG only passes the value to the ioctl.
I don't think this fix aligns with the existing testing program.

Please either change the test program or leave the ioctl logic unchanged.

linD026 (approved):

jserv:
Thank @applepwc for contributing!

PR #363, opened 2026-03-06
Add NULL-check platform_data before dereferencing in devicemodel_probe

EricccTaiwan:
dev->dev.platform_data is optional and not guaranteed to be set by the platform device registrar. Add a NULL guard and return -EINVAL to fail the probe cleanly instead of crashing the kernel with a NULL dereference.

---
## Summary by cubic
Add a NULL check for platform_data in devicemodel_probe to prevent a NULL dereference and kernel crash. If platform data is missing, log an error and return -EINVAL to fail the probe cleanly.

<sup>Written for commit 1d3a336b8324794779851cc81cb082885169ac7c. Summary will update on new commits.</sup>

### DIFF
--- a/examples/devicemodel.c
+++ b/examples/devicemodel.c
@@ -16,6 +16,11 @@ static int devicemodel_probe(struct platform_device *dev)
     struct devicemodel_data *pd =
         (struct devicemodel_data *)(dev->dev.platform_data);
 
+    if (!pd) {
+        pr_err("No platform data provided\n");
+        return -EINVAL;
+    }
+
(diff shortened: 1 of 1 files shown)


visitorckw:
To better justify this change, could you please clarify what are the specific conditions or execution paths that would trigger a NULL platform_data?

EricccTaiwan:
> To better justify this change, could you please clarify what are the specific conditions or execution paths that would trigger a NULL platform_data?

Added it out of habit from real driver patterns.
Happy to close if it's not appropriate.

